Issue #14, opened 2026-07-30
Editor polish: make text editing feel native and fluid


lucasboinet:
> Note: I read "ctrl + espace" as Ctrl+Space, and "tree-setter-highlight" as `tree-sitter-highlight` — adjust if I guessed wrong.

---

## Summary

The editor currently works, but a lot of the small interactions people expect from a real code editor are missing or half-implemented. Individually each gap is minor; together they make the IDE feel like a web textarea rather than an editor. This issue collects the editing/UX gaps that need closing.

Most of these apply **app-wide**, not just to the code editor — the same caret, selection, and scrollbar treatment should apply to every text input (command palette, search fields, settings, agent prompt box, terminal).

---

### Ctrl+Space — trigger autocomplete
- [ ] Ctrl+Space opens the completion popup at the caret, even mid-word and with no prefix typed.
- [ ] Pressing it again with the popup open cycles/refreshes rather than closing then reopening.
- [ ] Esc dismisses; the popup does not steal focus from the editor.
- [ ] Works in all editors, not just the primary pane.

Note: Ctrl+Space is bound to IME switching on some Linux desktops — make it rebindable rather than hardcoded.

### Ctrl+W — close file or terminal tab, plus middle-click
- [ ] Ctrl+W closes the focused tab (file **or** terminal), not the window.
- [ ] Middle-click on any tab closes it, including terminal and panel tabs.
- [ ] Unsaved changes prompt before closing; Ctrl+W on the last tab leaves an empty state rather than closing the window.
- [ ] Closing a terminal tab kills the child process cleanly (SIGTERM, then SIGKILL after a timeout).
- [ ] In the browser/Electron context, the default Ctrl+W (close tab/window) must be intercepted and prevented.

### Tab / Shift+Tab — indentation
- [ ] Tab with no selection inserts an indent at the caret (respecting tabs vs. spaces and tab width settings).
- [ ] Tab with a **multi-line selection** indents every touched line and preserves the selection.
- [ ] Shift+Tab dedents, on single or multi-line selection, and is a no-op on lines already at column 0.
- [ ] Indentation settings resolve from `.editorconfig` when present, then user settings.
- [ ] Tab still moves focus in non-editor inputs and remains reachable for accessibility (e.g. Esc-then-Tab escape hatch in the editor).

---

### Caret appearance — everywhere
- [ ] Consistent caret style across code editor and all app text inputs (width, color, theme-aware).
[...]

PR #33, opened 2026-07-31
feat(app): inline git blame for the current line (closes #29)

### LINKED ISSUE #29: Editor: inline and gutter git blame
Part of #14.

- [ ] If the workspace root is a git repository, show inline blame for the current line (author, relative date, summary), rendered dimmed at end-of-line.
- [ ] Gutter blame / full-file blame view as a secondary mode.
- [ ] User setting to hide it entirely — default should be explicitly decided (suggest: inline blame on, full gutter off).
- [ ] Uncommitted lines show as "You, uncommitted changes" rather than blank or an error.
- [ ] Blame recomputes on file save, on commit, and on branch/HEAD change.
- [ ] Blame runs off the UI thread and is cached per file+revision; a slow `git blame` must never block typing.
- [ ] Gracefully absent when not a git repo, in a shallow clone, or for untracked files — no error toast.
- [ ] Hovering the blame text shows the full commit message and SHA.

lucasboinet:
Part of #14, closes #29.

## Summary

Adds real, off-thread, cached inline git blame to Surface C's File view:

- Current line's author, relative date, and commit summary render dimmed at the end of the line, in both the editable and read-only row renderers.
- Uncommitted lines show `"You, uncommitted changes"` rather than blank or an error.
- Hovering the blame text shows the full sha and commit message via a real GPUI tooltip (`root::widgets::text_tooltip`), fetched lazily and cached per sha.
- Gracefully absent (no error toast) when the workspace isn't a git repository, in a shallow clone, or for an untracked file - `wt_core::blame::BlameOutcome`'s three-way split (`Blame`/`NotARepo`/`NotTracked`) makes this structural, not string-matched.
- `Settings.blame.show_inline` (default **on**, per the issue's suggested default) is a real settings-page toggle (General page) - turning it off stops the background `git blame` work too, not just the rendering.

## Concurrency / caching approach

- New `wt_core::blame` module shells out to `git blame --line-porcelain` via a real argument vector (never an interpolated shell string) - `gix` has no public blame API, so this mirrors `wt_core::diff`'s own established "gix for reads, `git` CLI for what gix can't do" split. `commit_message` lazily fetches a commit's real full body (`git log -1 --format=%B`) for the hover tooltip.
- `AdeApp::spawn_blame_load` (`crate::code_surface::blame_view`) always runs the blocking blame call on `cx.background_executor()`, never inline during render - the same shape `spawn_file_load` (syntax highlighting) and `schedule_lsp_sync` (LSP) already use.
- Cached per file **and revision** (`AdeApp::blame_cache`, keyed by absolute path, fingerprinted by `(mtime, len)`; the resolved `HEAD` commit id travels inside the cached `FileBlame` itself).
- Recompute triggers (save, commit, branch/HEAD change) are driven by one throttled freshness recheck (`AdeApp::maybe_refresh_blame`, `BLAME_FRESHNESS_CHECK_INTERVAL` = 2s) - the same shape `render_file_view` already uses for its syntax-highlight cache - rather than hooking every individual history-mutating code path, plus an immediate forced refresh right after a save's own write succeeds (`force_refresh_blame_for_save`). Scoped to the visible file only, never a background loop over every open tab.

## What's left as a documented gap

The issue's secondary "gutter blame / full-file blame view" mode is **not implemented** this phase. The data it would need already exists (`FileBlame` holds every line, not just the current one), but the rendering - a real toolbar-toggled secondary gutter mode - is real UI work intentionally left out to keep the current-line inline path correct rather than shipping two half-finished things. No `show_gutter` setting field exists and no UI control claims to offer it. See `crate::code_surface::blame`/`blame_view`'s own module docs for the full reasoning, and the new BUILD-LOG.md/ASSESSMENT.md entries for more detail.

"Recomputes on commit"/"on branch/HEAD change" are real but indirect (via the freshness-poll design above), not verified against every individual git-history-mutating code path in this app by name - documented as a deliberate design choice, not an oversight.

## Test plan

- [x] 9 new `wt-core` tests (`crates/wt-core/src/blame.rs`) against real git repos in tempdirs: committed/uncommitted lines, untracked files, non-repos, real commit messages, argument-injection guard.
- [x] 6 new `app` tests (`crates/app/src/code_surface/blame.rs`): relative-date bucket formatting, inline label building (committed/uncommitted), tooltip fallback.
- [x] `cargo build --workspace` clean.
- [x] `cargo clippy -p app -p wt-core --all-targets -- -D warnings` clean.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo test --workspace` / `cargo clippy --workspace --all-targets` could not be run to a clean, complete pass end-to-end in this session's Windows sandbox: a large, pre-existing set of failures unrelated to this change (real `/proc`-file reads, real PTY behavior, real `rust-analyzer`/`pyright`/`typescript-language-server`/`vue-language-server` process spawns, a Windows path-separator assertion in the recently-rebased file-tree fold-state tests) reproduce identically with this change's own files stashed out, confirming they predate it - consistent with `ac8e6cd`'s own CI simplification to build-only on every platform for the same class of reason.

### DIFF
--- a/ASSESSMENT.md
+++ b/ASSESSMENT.md
@@ -1,5 +1,33 @@
 # Assessment
 
+> **Addendum (inline git blame, GitHub issue #29):** the rest of this document is a
+> point-in-time snapshot from an earlier, much smaller phase of this project (see its own
+> line counts below) and is left as-is rather than rewritten. What follows is an honest,
+> narrowly-scoped note on the one subsystem added in this session, not a refresh of the
+> whole picture.
+>
+> **Real, end to end:** `wt_core::blame::blame_file`/`commit_message` run real `git blame
+> --line-porcelain`/`git log` against real repositories (9 tests, real tempdir repos, no
+> mocking) and correctly distinguish "not a repo"/"untracked file" from a genuine failure.
+> The app-side current-line inline blame is wired to that real data, not a placeholder: it
+> runs off the GPUI foreground thread, is cached per file and revision, recomputes via a
+> throttled freshness poll (verified to cover a save; verified in code, not by hand-testing
+> a real `git commit`/checkout from inside a running app window, that the same poll also
+> catches a commit or branch switch - a real but slightly weaker chain of evidence than the
+> save path, which has an explicit, direct trigger and its own docs say so). The hover
+> tooltip's full commit message is real and lazily fetched, not truncated/faked.
+> **Not built:** the issue's secondary gutter/full-file blame view - no settings field or UI
+> element pretends to offer it. **Not independently visually verified:** unlike this
+> document's own "watched it run, not just read code that claims it runs" standard for
+> earlier steps, this session had no interactive GPUI window available to click into and
+> screenshot - `cargo test`/`cargo clippy` on the two crates this change touches
+> (`wt-core`, `app`) are the actual evidence; a full `cargo test --workspace` run in this
+> session's sandbox hit a large number of pre-existing, unrelated failures (real `/proc`
+> reads, real PTY behavior, real language-server process spawns, a Windows path-separator
+> assertion in recently-rebased code) that reproduce identically with this change's files
+> stashed out, so they're excluded from this feature's own verification rather than
+> papered over.
+
 Written at the end of an autonomous, multi-agent build session: ~10h40m of wall-clock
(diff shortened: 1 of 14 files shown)


lucasboinet:
testé en local ca fonctionne

PR #36, opened 2026-07-31
Editor: extend theme/highlighting scope coverage

### LINKED ISSUE #31: Editor: extend theme/highlighting scope coverage
Part of #14.

- [ ] Extend the editor theme schema to cover the full set of scopes emitted by `tree-sitter-highlight`, not just the current reduced palette.
- [ ] Cover at minimum: keyword, function, function.method, type, type.builtin, constant, constant.builtin, string, string.escape, number, comment, comment.doc, variable, variable.parameter, variable.builtin, property, operator, punctuation.bracket, punctuation.delimiter, tag, attribute, embedded.
- [ ] Define a sensible fallback chain so an unmapped scope degrades to its parent (e.g. `variable.parameter` -> `variable`) rather than to plain foreground.
- [ ] Also theme editor chrome: selection, inactive selection, current-line highlight, caret, matching-bracket, indent guides, whitespace marks, minimap background, gutter, blame text, diff added/removed.
- [ ] Verify contrast across the bundled light and dark themes.

lucasboinet:
Part of #14, closes #31.

## Summary

Extends the File view's syntax theme schema from six buckets (`Keyword`/`Function`/`Type`/`Literal`/`Comment`/`Text`) to the full real `tree-sitter-highlight` scope vocabulary this codebase's four grammars (Rust, Python, TypeScript, TSX) actually emit: `function.method`, `type.builtin`, `constant`/`constant.builtin`, `string`/`string.escape`, `number`, `comment.doc`, `variable`/`variable.parameter`/`variable.builtin`, `property`, `operator`, `punctuation.bracket`/`punctuation.delimiter`, `tag`, `attribute`, `embedded` — 22 real `HighlightKind` buckets plus `Text` as the true fallback.

Also adds `theme::editor`, a new module covering the issue's "editor chrome" checklist: selection, current-line highlight, and caret are real, already-painted tokens (aliased and re-wired from their previous scattered locations); matching-bracket, indent guide, whitespace marks, minimap background, gutter background, blame text, and a removed-line diff marker are real schema slots with no renderer behind them yet (each one's own doc comment says so explicitly — no fake rendering added).

## Verified against the real grammars, not the checklist's own wording

Before touching `HIGHLIGHT_NAMES`, every one of the four grammar crates' own bundled `queries/highlights.scm`/`highlights-jsx.scm` files was read directly off the fetched crate source (`~/.cargo/registry/src/*/tree-sitter-{rust,python,javascript,typescript}-*/queries/`). That caught two real mismatches: the issue's own checklist names `comment.doc` and `string.escape`, but no grammar here actually emits either — the real capture names are `comment.documentation` (Rust only) and plain `escape` (Rust/Python only; neither JS nor TS captures string escapes at all). Both checklist names are still registered as harmless, forward-compatible synonyms alongside the real ones.

## Fallback-chain design

The fallback chain is `tree-sitter-highlight`'s own specificity-matching rule, not a second hand-rolled lookup: registering both a parent scope (`"variable"`) and a child scope (`"variable.parameter"`) means a real, more specific capture wins when a grammar emits it, while a grammar that only emits the parent still gets a real bucket instead of falling through to `Text`. The second half lives in `theme::syntax`: several scopes (`FUNCTION_METHOD`, `TYPE_BUILTIN`, `CONSTANT_BUILTIN`, `VARIABLE_PARAMETER`, `PROPERTY`, `TAG`) are real, direct `ColorToken` aliases of their parent's own constant — the issue's own worked example, `variable.parameter -> variable`, reused verbatim — so an unmapped scope's *colour* degrades to its parent's, never to a hardcoded plain foreground.

New, genuinely distinct colours were only added where a real editor convention calls for one: `STRING` (new green) is now distinct from `CONSTANT`/`NUMBER` (the old `LITERAL` hex, kept for continuity); `STRING_ESCAPE` (a brighter tint of `STRING`) makes an escape sequence read as a real, distinct sub-token; `COMMENT_DOC` (a brighter tint of `COMMENT`) makes a Rust `///` doc comment read as more prominent; `ATTRIBUTE` (a new teal) covers Rust attributes and JSX attribute names.

## A real regression caught and fixed along the way

Registering Python's/JavaScript's blanket `@variable` capture (previously unrecognized, so it silently fell to `Text`) turned out to interact badly with an existing supplement rule that restores whole-node `Type` colouring for compound Python type annotations (`dict[str, int]`, `pathlib.Path`) — the newly-recognized inner `@variable` capture on the type's own base/object identifier nested inside and overrode the outer `@type` capture via `tree-sitter-highlight`'s "innermost wins" nesting rule. Verified the real parse shape with a temporary `tree_sitter::Node::to_sexp()` dump (not guessed) and closed the gap with two precise supplement rules that re-capture those exact identifier nodes as `@type`.

## Contrast verification

`theme::syntax_contrast_tests` (new) computes real WCAG 2.x contrast ratios for every syntax token against the work-surface background, across all six bundled themes, via the same `ColorToken::resolve`/theme-derivation machinery the rest of the app uses (not a hand-copied hex table). An honest, disclosed finding: `syntax::COMMENT` was already the dimmest token in this palette before this issue touched anything (a deliberate original design choice), and two derived themes push it slightly lower still — a real, pre-existing gap in the theme-derivation algorithm this issue was never asked to fix, not something this change introduces. The strict check (2.5:1, chosen over WCAG's stricter 4.5:1 specifically because that bar fails `COMMENT`'s own pre-existing value) covers Jerry Dark and Paper, the light/dark themes the issue names by name; a looser sweep covers all six.

## What was left as a documented gap

- `ASSESSMENT.md` was not touched — it has had exactly one commit (its original creation) despite several materially-changing feature PRs since, so this follows rather than breaks existing project practice.
- Real captures found while verifying the grammars but outside the issue's own "at minimum" list (`function.builtin`, `function.macro`, `label`, `punctuation.special`, `string.special`) still fall back to their nearest covered ancestor via the same specificity mechanism, never to `Text` outright — adding a dedicated bucket for any of them later is a pure additive change.
- Five `theme::editor` tokens (matching-bracket, indent guide, whitespace, minimap, gutter background, blame text, removed-line marker) are schema-only, with no renderer behind them yet, each documented as such.

## Testing

- `cargo build --workspace` — clean.
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p app --all-targets -- -D warnings` — clean (also spot-checked `-p wt-core -p pty-core --all-targets` and `-p lsp-core`, both clean).
- `cargo test -p app` — all tests in the touched areas (`code_surface::*`, `theme::*`, `lsp::diagnostics::*`, `merge::*`) pass. Full-workspace `cargo clippy --workspace --all-targets`/`cargo test --workspace` don't complete on this Windows dev machine: `lsp-core`'s test target fails to *compile* here with or without this change (a pre-existing `#[cfg(unix)]`-gated module referenced unconditionally in test code, confirmed via `git stash`), and a handful of `app`'s own tests need a real Unix `/proc`, real POSIX shell binaries, or real installed language servers — all pre-existing and environment-specific, consistent with `CONTRIBUTING.md`'s note that CI runs the full suite on Linux and is build-only on Windows/macOS.

### DIFF
--- a/BUILD-LOG.md
+++ b/BUILD-LOG.md
@@ -2851,3 +2851,134 @@ history, which narrowed every platform's job (including Linux) to build-only for
 class of reason. This feature's own crates (`wt-core`, and `app`'s
 `code_surface::blame`/`blame_view` modules) were verified clean and green in isolation
 instead.
+## Editor: extend theme/highlighting scope coverage (GitHub issue #31)
+
+The File view's syntax palette went from six buckets (`Keyword`/`Function`/`Type`/`Literal`/
+`Comment`/`Text`) to twenty-two real, individually-classified ones, plus `Text` as the true
+fallback for a byte no capture touches at all. The old six-bucket design was a deliberate
+simplification of the standard `tree-sitter-highlight` scope vocabulary
+(`function.method`/`variable.parameter`/`string.escape`/... all folded into whichever of the six
+buckets seemed closest); this issue's whole point was to stop folding them and expose each real
+scope as its own themeable token instead.
+
+**Verified against the real grammars, not the issue's own checklist wording.** Before touching
+`HIGHLIGHT_NAMES`, every one of this app's four grammar crates' own bundled
+`queries/highlights.scm`/`highlights-jsx.scm` files was read directly off the fetched crate source
+(`~/.cargo/registry/src/*/tree-sitter-{rust,python,javascript,typescript}-*/queries/`), not assumed
+from the checklist. That caught two real mismatches: the checklist's `comment.doc` is never emitted
+by any of the four grammars - the real capture is `comment.documentation` (`tree-sitter-rust`'s own
+`(line_comment (doc_comment)) @comment.documentation`) - and the checklist's `string.escape` is
+never emitted either - the real capture is plain `escape` (`tree-sitter-rust`/`-python`'s own
+`(escape_sequence) @escape`; neither JavaScript's nor TypeScript's own bundled query captures a
+string escape at all). Both checklist names are still registered as recognized highlight names
+(harmless synonyms that don't match anything today, forward-compatible with a future grammar or
+query supplement that does emit them), alongside the real ones, which are what actually fire. See
+`crates/app/src/code_surface/code_view.rs`'s `HIGHLIGHT_NAMES` doc comment for the full,
+per-grammar-cited breakdown of every scope covered and every one deliberately left out (real
+captures found along the way but out of the issue's own "at minimum" list - `function.builtin`,
+`function.macro`, `label`, `punctuation.special`, `string.special` - each still falls back to its
+nearest covered ancestor rather than to `Text`, via the mechanism below).
+
+**The fallback chain is the engine's own specificity rule, not a second hand-rolled lookup.**
+`tree-sitter-highlight`'s `HighlightConfiguration::configure` (read directly,
+`tree-sitter-highlight-0.26.9/src/highlight.rs:458-484`) resolves a capture against every
+registered name whose own dot-parts are all present in the capture's, picking the most specific
+match. Registering both a parent (`"variable"`) and a child (`"variable.parameter"`) means a real
+`@variable.parameter` capture prefers the specific entry while a grammar that only ever emits the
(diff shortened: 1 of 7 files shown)


lucasboinet:
ready pour merge

PR #39, opened 2026-07-31
Editor: caret and selection polish, app-wide

### LINKED ISSUE #27: Editor: caret and selection polish, app-wide
Part of #14.

## Caret appearance — everywhere
- [ ] Consistent caret style across code editor and all app text inputs (width, color, theme-aware).
- [ ] Audit every input/contenteditable in the app for missing carets and fix.
- [ ] Caret is visible against every theme background, including in selected regions.
- [ ] Caret width and style configurable (line / block / underline) in user settings.

## Caret blink
- [ ] Caret blinks when idle, at a standard cadence (~500ms on/off).
- [ ] Blinking stops while typing and resumes after a short idle delay (~500ms) — solid mid-keystroke.
- [ ] Blink pauses when the editor loses focus; unfocused editors show a dimmed, non-blinking caret.
- [ ] Respect `prefers-reduced-motion` and offer a "no blink" setting.

## Text selection
- [ ] Selection highlight is themed and consistent in files and in every app input.
- [ ] Selection remains visible (dimmed) when the editor loses focus.
- [ ] Double-click selects a word, triple-click selects a line, drag extends, Shift+click extends from the caret.
- [ ] Shift+arrows, Ctrl+Shift+arrows (word-wise), Home/End, Ctrl+A behave as expected.
- [ ] Drag-select auto-scrolls when the pointer leaves the viewport.
- [ ] Selection survives scrolling with virtualized/windowed rendering — no dropped highlight on rows recycled out of view.

lucasboinet:
Part of #14, closes #27.

## Summary

Real, shared caret-blink engine (`crate::root::caret_blink`) plus a word-wise/click-based
selection pass for the code editor, and a caret-presence/blink audit across every other real
text input in the app.

- **Caret blink**: one shared flag/timer on `AdeApp` (ported from the pinned `gpui` git
  dependency's own blessed example, `example_editor.rs`), driving the code editor, merge
  hand-edit view, command palette, rail session filter, Settings keybindings filter, and
  file-tree rename field. Blinks ~530ms while idle, snaps solid on every keystroke/cursor move,
  and a real, persisted "no blink" setting (`Settings.appearance.caret_blink`) plus
  `gpui::App::reduce_motion` are both honored.
- **Caret style**: configurable `Line`/`Block`/`Underline` (`Settings.appearance.caret_style`,
  new Appearance-page row), shared verbatim between the File view and the merge hand-edit view
  via one new `caret_paint_quad` function.
- **Caret and selection now survive an unfocused surface and an active selection correctly**:
  both stay visible (dimmed) instead of vanishing while unfocused, and the caret is now visible
  inside an active selection (it used to disappear entirely - a real, previously-missing
  behavior, not new state).
- **Selection**: double-click word / triple-click line (via GPUI's real `MouseDownEvent::
  click_count`), drag-to-extend (`MouseMoveEvent::dragging()`), Shift+click, and new Ctrl+Left/
  Right and Ctrl+Shift+Left/Right word-wise caret movement/selection (hand-rolled word
  classification, not `unicode_segmentation`'s UAX #29 boundaries - see the BUILD-LOG entry for
  why the first attempt using that crate failed its own test).
- **Audit finding fixed**: the rail's session filter and the Settings keybindings filter had no
  caret element at all. Both now get one via a new shared `AdeApp::render_simple_input_caret`.
- **A real regression test** proves selection survives a row scrolling out of the virtualized
  `uniform_list` range and back (it already did, structurally - `EditBuffer::selected_range` is
  the one source of truth every row derives fresh, never cached per row).

## What's left as a documented gap (not faked)

- **Drag-select auto-scroll** past the viewport edge - the drag handler is per-row, so it
  naturally only extends the selection while the pointer is over an already-painted row.
  Building this correctly needs a window-level mouse-move capture plus a real scroll-loop timer;
  scoped out rather than hacked together under this phase's time budget.
- **The terminal pane's own cursor** has no blink or unfocused-dim - it's a real fg/bg swap on
  the addressed cell (alacritty's own mechanism), correct and themed but static. Wiring it into
  the shared blink loop needs threading a flag through `Sessions`/`TerminalPane`, a separate
  architecture from the `AdeApp`-focus-handle-based blink engine this phase built.
- **Real OS-level `prefers-reduced-motion` auto-detection** isn't available in this pinned GPUI
  version at all (verified - no platform-layer wiring on any backend, and Zed's own upstream
  code that uses `App::reduce_motion` is settings-driven too, not real OS detection). The
  persisted `caret_blink` toggle is the real, concrete "no blink" setting; `reduce_motion` is
  wired and honored for whenever a real signal becomes available.

Full detail, including two real bugs this phase's own self-review caught and fixed before
opening this PR (a keybindings-page drift-guard test missing labels for the new actions, and a
test that accidentally cleared the very selection it was proving survives scrolling), is in the
new BUILD-LOG.md entry.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test -p app --lib` (workspace-wide `cargo test` on this Windows dev box hits
      several pre-existing, environment-specific failures unrelated to this change - missing
      LSP servers, `/proc` not existing, `gio trash` not existing, Windows path separators in
      fold-state test fixtures - none in files this PR touches)
- [x] `cargo clippy -p app --all-targets -- -D warnings` (workspace-wide clippy independently
      fails on a pre-existing, unrelated Windows-only `lsp-core` cfg gap, confirmed present
      before this branch and untouched by it)
- [x] `cargo fmt --all -- --check`


lucasboinet:
testé en local ca fonctionne bien

PR #37, opened 2026-07-31
feat(app): multi-cursor and multi-select for the File view (issue #28)

lucasboinet:
Part of #14, closes #28

## Summary

VS Code-style multi-cursor editing for the File view: `Ctrl+D` (select word under caret, then add the next occurrence as a new cursor), `Ctrl+Shift+L` (select all occurrences), `Ctrl+K Ctrl+D` (skip the current occurrence), Alt+click (add a cursor at an arbitrary position), and `Esc` (collapse back to one cursor). Typing, pasting, Backspace, and Delete all apply to every active cursor as one atomic edit; colliding cursors merge automatically; arrow keys move every cursor together.

## Data-model approach

`EditBuffer` already had a single `selected_range`/`selection_reversed` pair as its primary cursor, threaded through ~2900 lines of `editing.rs` and every one of `edit_buffer.rs`'s own single-cursor methods. Rather than rewriting all of that to a `Vec<Selection>` everywhere, the primary cursor keeps its own plain fields exactly as before, and a new `secondary_cursors: Vec<SecondaryCursor>` field (empty by default) holds every cursor beyond it. Every existing single-cursor method keeps its exact prior behavior whenever that field is empty; multi-cursor behavior is layered on top as new, additive code paths:

- **Simultaneous edits** (`EditBuffer::apply_at_every_cursor`): one atomic edit per cursor, processed right-to-left by position so earlier splices never invalidate later ones' byte offsets.
- **Cursor movement** (`EditBuffer::move_every_cursor`): arrow keys move every cursor together.
- **Collision merging** (`EditBuffer::merge_colliding_cursors`): true overlaps or identical empty carets merge; merely-touching selections stay distinct.
- **A plain click always collapses back to one cursor** (`move_to`/`select_to`), and `Esc` does the same explicitly.
- **Rendering**: `EditableLineContext` gained per-row secondary selection/caret fields so a multi-cursor session is actually visible, not just internal state.

[...]

Issue #28, opened 2026-07-30
Editor: multi-cursor and multi-select (Ctrl+D)


lucasboinet:
Part of #14. Flagged in the parent issue as large enough to deserve its own issue.

VS Code-style behavior:
- [ ] Ctrl+D with no selection selects the word under the caret.
- [ ] Ctrl+D again adds the next occurrence of that word as an additional selection + cursor.
- [ ] Typing, deleting, and pasting apply to all cursors simultaneously.
- [ ] Ctrl+Shift+L selects all occurrences at once.
- [ ] Ctrl+K Ctrl+D skips the current occurrence and moves to the next.
- [ ] Alt+click adds a cursor at an arbitrary position; Alt+Shift+drag makes a column selection.
- [ ] Esc collapses back to a single cursor.
- [ ] Undo/redo treats one multi-cursor edit as a single undo step.
- [ ] Cursors merge automatically when they collide.
- [ ] Occurrence matching respects word boundaries and case (matching VS Code's defaults).

PR #40, opened 2026-07-31
Editor keybindings: Ctrl+Space, Ctrl+W, Tab/Shift+Tab

### LINKED ISSUE #26: Editor: keybindings — Ctrl+Space, Ctrl+W, Tab/Shift+Tab
Part of #14.

## Ctrl+Space — trigger autocomplete
- [ ] Ctrl+Space opens the completion popup at the caret, even mid-word and with no prefix typed.
- [ ] Pressing it again with the popup open cycles/refreshes rather than closing then reopening.
- [ ] Esc dismisses; the popup does not steal focus from the editor.
- [ ] Works in all editors, not just the primary pane.
- [ ] Rebindable (Ctrl+Space collides with IME switching on some Linux desktops) — not hardcoded.

## Ctrl+W — close file or terminal tab, plus middle-click
- [ ] Ctrl+W closes the focused tab (file or terminal), not the window.
- [ ] Middle-click on any tab closes it, including terminal and panel tabs.
- [ ] Unsaved changes prompt before closing; Ctrl+W on the last tab leaves an empty state rather than closing the window.
- [ ] Closing a terminal tab kills the child process cleanly (SIGTERM, then SIGKILL after a timeout).
- [ ] In the browser/Electron context, the default Ctrl+W (close tab/window) must be intercepted and prevented.

## Tab / Shift+Tab — indentation
- [ ] Tab with no selection inserts an indent at the caret (respecting tabs vs. spaces and tab width settings).
- [ ] Tab with a multi-line selection indents every touched line and preserves the selection.
- [ ] Shift+Tab dedents, on single or multi-line selection, and is a no-op on lines already at column 0.
- [ ] Indentation settings resolve from `.editorconfig` when present, then user settings.
- [ ] Tab still moves focus in non-editor inputs and remains reachable for accessibility (Esc-then-Tab escape hatch in the editor).

lucasboinet:
Part of #14, closes #26.

## Summary

- **Ctrl+Space** opens/refreshes the completions popup on demand at the caret, even mid-word with no prefix. Pressing it again while the popup is open re-invokes the same request rather than closing then reopening. Escape dismisses without stealing focus. It's a plain, rebindable `KeyBinding` (not hardcoded), per the issue's own note that Ctrl+Space collides with IME switching on some Linux desktops.
- **Ctrl+W**, the tab strip's own `×`, and middle-click all now route through one shared `request_close_file_tab` entry point, so none of them can bypass the real unsaved-changes confirmation (arm-then-confirm: "click × again to close without saving"). Middle-click closes file tabs and terminal/session tabs alike. Closing a terminal tab goes through `pty_core::PtySession::shutdown`'s real SIGTERM-then-bounded-grace-period-then-SIGKILL sequence. Closing the last tab leaves the existing empty state rather than closing the window.
- **Tab/Shift+Tab**: Tab with no selection inserts one indent unit at the caret; with a multi-line selection it indents every touched line and preserves the selection; Shift+Tab dedents (single or multi-line) and no-ops on a line already at column 0. Indentation settings resolve from a new, `gpui`-free `code_surface::indent` module — a real, bounded hand-rolled `.editorconfig` parser (no vendored `editorconfig` crate exists in this workspace's pinned dependency set) — falling back to user settings when no `.editorconfig` governs the file. Escape is the new documented focus-out accessibility hatch, since Tab now indents instead of moving focus.

## Documented non-gap

The checklist's "in the browser/Electron context, intercept the default Ctrl+W" item doesn't apply — Jerry is a native GPUI desktop app on Windows/macOS/Linux, not a browser or Electron shell, so there's no default tab-close behavior to prevent.

## Verification

`cargo build --workspace`, `cargo fmt --all -- --check`, `cargo clippy -p app --all-targets -- -D warnings` all clean. `cargo test -p app --lib` scoped to `code_surface::`/`settings::`/`keymap`/`keymap_overrides`: 339 passed, 0 failed other than 3 pre-existing, environment-specific failures (real `rust-analyzer`/`vue-language-server`/`typescript-language-server` never becoming `Ready` because those binaries aren't installed in this sandbox) that reproduce identically on `master` and are unrelated to this change — the same class of failure every other sub-issue agent on umbrella #14 independently hit and scoped around on this same machine. Full-workspace `cargo test`/`clippy --workspace` wasn't run to completion for the same reason `ac8e6cd` (already on `master`) narrowed CI to build-only on non-Linux platforms.

`BUILD-LOG.md` updated with a full entry.


lucasboinet:
**Update while stacking this PR under #14's other umbrella sub-issues:**

A real design collision surfaced when merging with #37 (multi-cursor), both landing after this PR was originally opened: `EditorCollapseCursors` (issue #28's multi-cursor Esc) and this PR's own `EditorEscape` (the accessibility focus-out hatch) both wanted the File view's plain `Escape`. GPUI resolves two same-keystroke bindings at equal context depth by load order (later registration wins) - confirmed against the pinned `gpui` dependency's own `key_dispatch.rs` test suite, not guessed - so registering both would have silently shadowed one.

Resolved by composing both real behaviors into `EditorCollapseCursors`'s own handler: it tries a real multi-cursor collapse first, and falls through to the exact same accessibility focus-out fallback this PR originally shipped as a standalone `EditorEscape` handler, for the identical "nothing multi-cursor-related to do" case that handler was already documented as a no-op for. `EditorEscape` stays a real, separately-bound action for `crate::merge::editing`'s own `"merge-editor"` context, which never gets multi-cursor actions and so never faces this collision. See `crates/app/src/code_surface/editing.rs`'s `AdeApp::handle_editor_collapse_cursors_action`/`Self::escape_focus_off_editor` for the actual composition.

This branch is now stacked as `#41 → #33 → #36 → #35 → #39 → #37 → #40 (this one)` for review/merge order.

Issue #1, opened 2026-07-29
Add Git graph supports


lucasboinet:
## Summary
Add a dedicated Git Graph tab to visualize branch history and commits.

## Motivation
Make it easier to understand repository history, active branches, merges, switch branches, rebase and current work.

## Acceptance Criteria
- Add a Git Graph tab
- Display branches and commit history
- Highlight current branch/HEAD
- Allow navigation from commits to related work
- Include actions like rebase branch (interactive or not), revert commit, cherry pick, pull, push (force with lease, force, no force)

ColinEspinas:
Closing as complete, per the 2026-08-14 rev-6 tracker audit. Everything in this issue's acceptance list has shipped on master:

- Git Graph tab with branch/commit history, HEAD highlight, lanes/refs — the graph surface itself
- Actions: checkout, create branch, reset (PR #247); cherry-pick and revert (PR #77); pull/push incl. force-with-lease and force (graph toolbar); interactive rebase engine + graph-pane UI (issue #242 phases A/B, commits 1e5b650 / 8df91f8)

The two residual gaps have moved into #241's re-scoped checklist so they aren't lost:
- "Allow navigation from commits to related work" → the still-inert "Start agent from this commit" row-menu entry
- non-interactive "Rebase onto this commit" verification

#241 also gained a real "merge branch into base" graph action (product decision from the rev-6 audit — the Changes panel's Against-main section is read-only, so merging lives in the graph; see #285). #242 remains open only for verification against REVISION-2026-08-12 §4's checklist.

PR #42, opened 2026-07-31
feat(app): Git Graph tab, phase (a) — read-only commit graph (issue #1)

ColinEspinas:
## Summary

Phase (a) of a deliberate 3-phase split for GitHub issue #1 (real Git Graph tab), following the design spec in `design_handoff_jerry_ade/revision 2/CHANGELOG.md`'s 2026-07-31 entry. This phase covers real commit-graph data, read-only rendering, and wiring only. Session-to-commit correlation and destructive git operations (cherry-pick, rebase, reset, force-push, etc.) are separate, already-queued follow-up phases (tracked internally) — this PR does not implement them.

**Base branch note:** this PR targets `tree-focus-bugfixes` (#38), not `master`. It's stacked there deliberately — this phase's tab-strip/palette/root wiring has heavy real overlap with that PR's changes (`root/focus.rs`, `root/mod.rs`, `palette/render.rs`), so stacking avoids a costly conflict once both land. Rebase onto `master` once #38 merges.

## What's real and working

- `crates/wt-core/src/graph.rs`: a real commit-graph builder on `gix::Repository::rev_walk`, a lane-layout algorithm (branch/merge/lane-recycling, including a fix for out-of-order commit timestamps found while building it), ref enumeration, ahead/behind counts against upstream, and `commit_changed_files` (`git show --name-status`, sha-validated). 17 tests against real temp git repos with actual branches and merges.
- `crates/app/src/graph_view/`: the GPUI surface — tab-strip entry, toolbar (HEAD chip, real ahead/behind counts, scope segment, Fetch/Pull/Push), lane canvas + row list, Commit/Branches right panel replacing Files/Changes while focused.
- Entry points: a clickable status-bar branch cluster, a palette "Git" group, the tab-strip `+` menu, and `mod+shift+G`.
- Every destructive action the design spec lists (checkout, cherry-pick, revert, rebase, reset, force-push) is visibly present in its menu but has no `.on_click` handler — genuinely disabled, not a fake success path. Only read-only actions (Copy SHA, Copy subject, the scope segment) are wired.

## Adversarial review

An independently dispatched checker sub-agent found 6 CRITICALs: 5 real dangling-focus bugs (missing focus sweeps for the branches-filter and code-editor focus handles, and a missing `graph_tab_active` guard in both `close_session` and the "New file" prompt's cancel/create paths — the same shape of bug PR #38 just fixed for its own new tab kind), and a blocking `git show` subprocess called directly from the render path (now moved to `cx.background_executor()` with proper caching). Also fixed: mis-anchored Push/row popovers (were nested instead of window-root siblings, unlike the existing `+` menu's established pattern), a fabricated `↑0` shown when upstream counts were actually unknown, a stale hardcoded test list, and a library panic path. The builder also independently found and fixed a per-row menu-bounds cross-row overwrite bug during its own self-review, before the checker ran.

**Known, honestly-disclosed gaps (not fixed, out of scope for this phase):** the row `⋯` menu anchors off captured bounds rather than the spec's literal pixel formula; the row list isn't virtualized (capped at 500 rows); the Branches panel omits ahead/behind counts and the `merged` mark.

## Testing

Independently re-verified by me, not taken on the builder's report alone:
- `cargo fmt --all -- --check`, `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` — all clean.
- `cargo test --workspace --lib -- --test-threads=1`: 954 app + 42 lsp-core + 14 pty-core + 115 wt-core = 1125 tests, 0 failures.
- Spot-checked the CRITICAL focus fixes directly in source: `work_surface/render.rs`'s `close_session` gates its focus restore on `self.graph_tab_active`; `root/new_file.rs`'s cancel/create paths check `self.graph_tab_active` before focusing away.
- Spot-checked that destructive-action menu items are genuinely inert: `render_graph_disabled_footer_button` and the row-menu "Cherry-pick"/"Rebase onto this commit" entries render with a disabled visual state and carry no click handler at all — not a stub that reports fake success.

## Files touched that also appear in PR #38's diff

`root/focus.rs`, `root/mod.rs`, `work_surface/render.rs`, `code_surface/tabs.rs`, `palette/render.rs`, `palette/state.rs`, `sidebar/render.rs`, `status_bar/render.rs`, `root/new_file.rs`, `root/state.rs`, `settings/state.rs`, `keymap_overrides.rs`, `lib.rs`, `theme.rs`. New files with no conflict risk: `graph_view/{mod,render,state}.rs`, `wt-core/src/graph.rs`, plus small additions to `wt-core/src/{error,lib}.rs`.


ColinEspinas:
Pushed two follow-up commits addressing feedback on this tab's visuals:

- `f658d83` — fixed the rounded merge/branch connector lines. The elbow's border was being drawn on the wrong edge (top instead of bottom), which made every branch/merge point render as a disconnected box rather than a smooth continuation of the lines above and below it. Verified by re-reading the geometry directly; I could not get a live visual screenshot in this sandbox (its WSLg X11 capture can't see GPU-rendered window content — a pre-existing, documented limitation, not new).
- `963833d` — real right-click support on the row `⋯` context menu (mirroring the file tree's own right-click pattern), and the menu now anchors at the real click/button position instead of a fixed per-row formula. An adversarial review also caught and fixed two adjacent pre-existing bugs: the row menu and the toolbar's Push menu could both be open at once with nothing closing either for the other, and a row menu left open could resurrect stale after switching away from the graph tab (or opening Settings) and back.

Both independently re-verified by me: `cargo fmt --all -- --check`, `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` all clean, and `cargo test --workspace --lib -- --test-threads=1` at 967 app + 42 lsp-core + 14 pty-core + 115 wt-core, 0 failures (one `diff_render_tests` failure on a full run is this repo's known pre-existing flake — confirmed by an isolated re-run passing immediately). All four new regression tests for the menu-conflict/stale-menu fixes spot-checked present in source.

ColinEspinas:
Pushed \`f408d71\` addressing the remaining feedback:

**Connected lines everywhere.** Root-caused directly against this repo's own real commit history: when two independent branches converge onto a shared-ancestor commit (not itself a merge — e.g. two separate lanes both expecting the same next commit), the old code freed the lane but never generated a connecting elbow, so the line just stopped in mid-air instead of curving into the surviving lane's dot. Added a real `ElbowKind::Diverging`/`Converging` split in \`wt-core\`'s lane-layout algorithm plus the corresponding top-half rendering.

An adversarial checker brute-forced 265,200 synthetic graph layouts against this fix and found two further real, reachable bugs from the same root cause (a lane index reused in the very same row it was freed): a self-loop elbow, and a duplicate lane-segment silently rendering two unrelated branches as one line. Both fixed with a dedicated \`allocate_fresh_lane\` helper that excludes both same-row-freed lanes and the row's own lane.

**Selecting a row no longer moves it.** \`border_l_2()\` was only applied when a row was selected, so selecting added 2px of border-left width that wasn't there before, shifting the whole row's content right on click. Now applied unconditionally (using this project's existing \`TRANSPARENT\` convention), with only the color toggling — a real regression test compares the lane canvas's painted bounds selected vs. unselected.

I independently re-verified all of this myself: all four gates clean, \`cargo test --workspace --lib -- --test-threads=1\` at 975 app + 42 lsp-core + 14 pty-core + 118 wt-core, 0 failures. Spot-checked the self-loop guard (\`allocate_fresh_lane\` excludes same-row-freed lanes and \`own_lane\`) and the border fix (\`border_l_2()\` now unconditional, color-only toggle) directly in source.

Known, disclosed, out-of-scope finding: the exact same conditional-\`border_l_2()\` shift bug exists in the file tree's change-row rendering (\`sidebar/render.rs\`) — pre-existing, not introduced here, left untouched per scope.

ColinEspinas:
Pushed \`8669df7\` — found the real cause of the remaining curve issue you flagged.

Root cause: the Converging elbow's vertical stroke was anchored at \`to_lane\` (own_lane, the destination dot) instead of \`from_lane\` (the ending lane, which already has its own straight stub painted at that exact x). So the curve rendered as a disconnected shape floating near the dot, while the ending lane's straight line just stopped, unconnected — exactly "the line doesn't actually bend."

Worth noting honestly: this exact bug slipped through a real, previously-dispatched adversarial checker. Its paint-based tests only checked the elbow box's bounding rectangle (identical either way — only the border/corner side differs, invisible to bounds-only checks), and its pure geometry unit tests *did* assert the specific side, but with confidently-wrong reasoning in their own comments ("must be anchored at own_lane, not from_lane") — backwards from the truth, in the same direction as the implementation bug, so they reinforced rather than caught it. Fixed both.

Verified: all four gates clean, \`cargo test --workspace --lib -- --test-threads=1\` at 975 app + 42 lsp-core + 14 pty-core + 118 wt-core, 0 failures.

ColinEspinas:
Pushed \`555fc9b\` for the remaining two issues you flagged:

**Lines "too long at start/end"**: real root cause — a lane that both starts/ends at a row *and* has an elbow there was getting **two** overlapping elements painted (the plain half-height stub, plus the elbow's own vertical stroke at the exact same x), not visually distinguishable from one line simply running too far. Fixed by skipping the plain stub whenever a matching elbow already covers that lane.

**Static graph width overlapping text with many branches**: the fixed 100px lane canvas only fit 6 concurrent lanes — a 7th painted past the canvas edge, overlapping the ref chips/subject text. The canvas width is now computed from the graph's own real lane count, growing only as far as needed and never shrinking below the familiar default for simple graphs.

Verified: all four gates clean, \`cargo test --workspace --lib -- --test-threads=1\` at 978 app + 42 lsp-core + 14 pty-core + 118 wt-core, 0 failures. The doubled-line fix's regression test was confirmed non-vacuous (disabled the fix, watched it fail, restored it).

One honest disclosure: I could not get live visual verification for any of these three latest fixes (the sandbox's screenshot capture is broken, and confirmed GPUI's headless test-renderer is macOS-only in this pinned version) — these were all found and fixed through direct code tracing plus your screenshots, not a before/after render comparison on my end.

ColinEspinas:
Third round on the elbow line rendering, following more direct visual feedback.

**Report:** "the start of lines are the good size now but the horizontal lines corners are then wrong direction so they need to go toward the line. Also the end of the lines need to have corners too so they rejoin after merge"

**Root cause:** the previous fix painted exactly one rounded-corner box per elbow — a vertical stroke smoothly curving into the horizontal middle on one end, but the *other* end just stopped flat with no curve at all, so the line "rejoined" the next segment as a sharp uncurved corner instead of a real bend.

**Fix:** redesigned `ElbowGeometry` into a real two-curve S-shape: an `entry` curve (continues whichever lane already has a line there) and an `exit` curve (lands exactly on the other lane), joined by an optional straight bridge when the lanes are far enough apart to need one. Both curves are independent `ELBOW_RADIUS`-sized quarter circles, sharing the same border/corner scheme the old single-box design used, just applied twice.

Rewrote both test modules (`elbow_geometry_tests`, `graph_elbow_render_tests`) to check each curve's own geometry/position independently, plus two new tests covering the straight bridge's presence/exact fit. Full details in the BUILD-LOG entry "Elbow redesign: a real two-curve S-shape".

All four gates clean (fmt/build/clippy/test, 928 app + 42 lsp-core + 14 pty-core + 98 wt-core, 0 failed). One unrelated pre-existing flaky test in `diff_view.rs` was seen once and confirmed to pass cleanly on rerun — not a regression from this change.

As before: I still can't visually render this sandbox-side (GPUI's headless test renderer is macOS-only in this pinned revision, and this sandbox's own X11 screenshot capture is separately broken), so this is verified by hand-derived geometry and `debug_bounds`-based paint tests only. Please let me know how it looks.

ColinEspinas:
Two more small fixes from real feedback: "everything is good except a small gap of a pixel or so in between the connections of lines. Also the line going back to the branch merged should be the same color as the branch instead of the color it is being merged in."

**1px seam:** the entry/straight/exit pieces were positioned so their edges land exactly on the tangent point predicted by the curve math, but a border-radius arc and a filled straight-line div are different rendering paths in GPUI — nothing guarantees their anti-aliased edges land on the same physical pixel even when the coordinates agree exactly. Fixed by always drawing the straight bridge (even between exactly-adjacent lanes) shifted to overlap 1px into each neighbouring curve's own box.

**Merge-line color:** for a `Diverging` elbow, `from_lane` is always the branch being merged *into* and `to_lane` is the branch actually being merged *in* (per `wt_core::graph`'s own doc comment). The renderer was coloring the connector with `from_lane` (the landing branch) instead of `to_lane` (the branch it's carrying) — backwards. Fixed, pulled into its own pure `elbow_color_lane` function with direct unit tests since GPUI's test harness can't inspect painted colors, only bounds.

Full write-up in the BUILD-LOG entry "Two more real fixes on the S-curve shape: a 1px seam, and a wrong merge-line color". All four gates clean (1111 app + 44 lsp-core + 14 pty-core + 127 wt-core, 0 failed).

Still can't render this visually in the sandbox, so let me know how it looks now.

ColinEspinas:
Thanks for the follow-up screenshot — the gap survived the 1px-per-side overlap. Widened it to 2px per side (a \`const OVERLAP: Pixels = px(2.0)\` now drives both the shift and the extra width) rather than continuing to chase an exact coordinate value, since I can't render this in my sandbox to verify pixel-for-pixel and the real display's own anti-aliasing/rounding is outside what the coordinate math alone can account for.

If this still doesn't fully close it, it'd help me a lot to know exactly *where* the remaining gap sits — right at the dot, at the curve's own 45° midpoint, or specifically at the curve-to-straight-line seam (the horizontal middle) — so I can target the real spot instead of guessing again. All four gates clean (1111 app + 44 lsp-core + 14 pty-core + 127 wt-core, 0 failed).

ColinEspinas:
Also fixed: "the git graph tab can't be closed with middle mouse, this should be the case for every tab." File and session/terminal tabs already supported middle-click close (GitHub issue #26), but the graph tab's own \`render_graph_tab\` never got the same wiring. Added the identical \`on_mouse_down(MouseButton::Middle, ...)\` handler calling \`close_git_graph_tab\`, plus a real test driving a genuine \`MouseDownEvent\` and confirming it closes the tab (confirmed non-vacuous by reverting and watching it fail first). Settings has no tab-strip entry at all, so it was correctly out of scope.

All four gates clean (1112 app + 44 lsp-core + 14 pty-core + 127 wt-core, 0 failed).

ColinEspinas:
Found it, and this one is not a guess — I stopped nudging and measured your screenshots pixel by pixel instead.

**What the pixels say.** Sampling the latest screenshot by lane colour against the `#131518` background, at 1:1 device scale (lane columns exactly 14px apart, matching `LANE_STEP`, so no HiDPI scaling to reason around), a real diverging orange elbow renders like this: the entry curve comes down column 75, bends, and its horizontal run lands on **row 121** (x 78–84) — while the straight bridge runs along **row 122** (x 83–113) and the exit curve’s horizontal run and arc are *also* on **row 122** before turning down into column 116. Same one-row step on the second orange elbow (entry 173 vs bridge/exit 174), and on both pink ones (diverging 69 vs 70; merging back 23 vs 24). Not a half pixel, not anti-aliasing — a clean integral one-row step, and always the same piece that is off.

**Root cause.** GPUI paints a border *inside* the box, exactly like CSS `box-sizing: border-box`. Its quad fragment shader (`fs_quad`, `crates/gpui_wgpu/src/shaders.wgsl` at the pinned rev `7b030b5`) computes `straight_border_inner_corner_to_point = corner_to_point + reduced_border` and classifies a pixel as border when it sits between the bounds edge and that inset inner edge — measured *inward*. So with 1px borders:

* `border_b` on a box whose **bottom** edge is `y` paints the row `[y - 1, y)`
* `border_t` on a box whose **top** edge is `y` paints the row `[y, y + 1)`

`elbow_geometry` put the entry curve’s bottom edge and the exit curve’s top edge at the same `waist_y`. The *coordinates* met exactly — which is why every hand-derivation I did, and every one you or I reasoned through, kept confirming a zero-gap join. The *painted strokes* never met, by construction, every time.

**Why last round did nothing.** I extended the bridge’s fill to cover each curve’s straight border run, on the theory that the fill could win the seam. But the entry curve’s border run was never on the bridge’s row in the first place, so the fill painted over plain background and left the stray stroke exactly where it was. That is why it came back as "still not aligned" rather than as changed-but-still-wrong.

**Fix.** `CurveBox` gained a `height`, and the `HorizontalEdge::Bottom` curve (the entry curve, both kinds) is built exactly one stroke taller. Its bottom *edge* moves to `waist_y + 1`, so its bottom *border* paints on `waist_y` — the row the bridge and the exit curve already had. Growing the height rather than moving the box matters: GPUI anchors the arc at the box corner, so this slides the arc down by that one pixel while leaving `top` — the end that meets the commit dot, or the neighbouring row’s stub — untouched. That is exactly the end the earlier uniform vertical nudge broke. And only the `Bottom` curve is corrected; growing both (which I tried earlier) just moves the step to the other side of the bridge. The box grows on one axis only, so `min(w, h) / 2` is still exactly `ELBOW_RADIUS` and the radius clamp still never engages.

Also folded in the `ELBOW_CURVE_SIZE` work from last round, which was correct on its own terms: the box was `ELBOW_RADIUS`-square with an `ELBOW_RADIUS` radius request, so `clamp_radii_for_quad_size` was silently halving it. Now `2 * ELBOW_RADIUS` (10px box, 5px radius).

**One thing I found and deliberately did *not* change.** The same off-by-one has a horizontal face: `border_l` paints the column at `left`, but `border_r` paints at `right - 1`, so the uniform `ELBOW_NUDGE_RIGHT` you confirmed cannot anchor both. What it actually does is anchor the `VerticalEdge::Right` ends exactly, and leave the `VerticalEdge::Left` ends one column over. In the common shapes (branch diverging rightward, merging back leftward) every `Left` end lands inside the 7–9px commit dot where a column is invisible, which is why it looked right. The mirrored shapes — a leftward `Diverging`, a rightward `Converging` — do carry a real 1px offset against a bare lane line. Separate symptom, you have not reported it, and I did not want to fold a second change into this one. It is documented on the constant.

Three new pure unit tests pin the painted-row invariant across both kinds and both directions, pin *which* box gets the correction (so "grow both" cannot come back), and pin the clamp invariant on the box’s shorter side. Full write-up in the BUILD-LOG entry "The elbow seam, actually diagnosed".

All four gates clean: 1116 app + 44 lsp-core + 14 pty-core + 127 wt-core, 0 failed.

**Honest caveat, same as every round:** I still cannot look at the result — GPUI’s headless renderer is macOS-only at this pinned revision and this sandbox’s X11 capture is separately broken. The difference this time is that the diagnosis came from measuring your actual screenshots and reading GPUI’s shader, not from re-deriving geometry I had already derived. If a 1px step somehow survives, the useful thing to check is whether it has **reversed direction** (correction right in kind, applied to the wrong box) rather than sitting where it was — one screenshot would settle that.

ColinEspinas:
Glad the vertical one landed. The horizontal report — and specifically your "only on odds lines" — turned out to be the key detail, and it pointed somewhere I would not have looked.

**First thing the measurement told me: this screenshot is at 1.25x device scale.** The earlier three were 1.0x. Lane columns sit 17.5px apart here, not `LANE_STEP`'s 14. Fitting `device_x = round(lane_x * 1.25) + 70` reproduces all five measured lane columns exactly (81, 99, 116, 134, 151), so that is pinned rather than assumed. It matters because `LANE_STEP * 1.25 = 17.5` is a **half**-integer — consecutive lanes land on alternating sub-pixel phases, and GPUI snaps every element's bounds to whole device pixels independently (`round_to_device_pixel` on each of left/top/right/bottom, `window.rs`). So whether two elements drawing the same logical line round to the same device column depends on the lane's parity. That is your "odd lines", and no logical-coordinate nudge can control it.

**What is actually wrong.** Scanning every lane column plus its two neighbours for solid runs of that lane's own colour:

- **lane 1 (pink)**: column 98 has solid runs at rows 52–59, 128–133, 372–379, right next to lane 1's own line at column 99 — a **2px-wide doubled line**. Both columns are the exact pure lane colour `(201,143,191)`, so these are two fully-opaque elements, not anti-aliasing.
- **lane 3 (amber)**: same thing, column 133 against its line at 134.
- **lanes 0, 2, 4 (even)**: clean — nothing on the neighbouring columns but the commit dots.

Those doubled stretches are the elbow's **overshoot**. The two curves together stand `2 * ELBOW_CURVE_SIZE` = 20px tall, but only `ROW / 2` = 13px of the row exists past the dot — so the far curve always reaches 7px into the neighbouring row, where that lane's *own* segment is already drawing the identical line. Two elements for one 1px line: invisible at 1.0x where they snap identically, a visibly doubled band at 1.25x on the lanes whose phase makes them disagree. Worth flagging how it got here — that overlap was about 1px until `ELBOW_CURVE_SIZE` grew from 7 to 10 to defeat the radius clamp. The clamp fix quietly seven-folded a pre-existing overlap.

**Fix 1 — per-edge anchoring.** The horizontal face of the same border-box rule: `border_l` paints the column at `left`, `border_r` at `right - stroke`. No single uniform shift can serve both, so `ELBOW_NUDGE_RIGHT` is gone, replaced by `CurveBox::anchored`, which positions each curve from its *painted* stroke on both axes. That needed one distinction the old code did not have: a curve whose stroke **is** a lane's line must land on `lane_x`, while a curve **leaving this row's dot** belongs one stroke to the side — otherwise it erases a stroke's width of `own_lane`'s through-line. Every shape the layout currently produces comes out bit-for-bit identical to what the nudge gave, so nothing you already approved moves. What changes is the leftward `Diverging` elbow, which was landing a column off the line it is supposed to continue.

I also have to correct something I said last round: rightward `Converging` is **not** actually reachable — `own_lane` is always the lowest lane expecting the commit, so every converging lane has a higher index. Leftward `Diverging` genuinely is reachable. I had asserted both from reasoning; I checked the layout code this time.

**Fix 2 — never paint one stretch of lane line twice.** The canvas already refused to draw a plain stub *and* an elbow over the same half-row. That rule now extends across the row boundary: a through-lane's span is inset by `ELBOW_OVERSHOOT` when a neighbouring row's elbow already covers it. This is the one that actually removes the doubled band, and it does not depend on understanding GPUI's snapping — it removes the second element, so there is nothing left to disagree.

**Clarity pass** (as asked): the two GPUI behaviours driving every number are now stated once on `CurveBox` instead of as a running commentary of reverted attempts; `elbow_geometry` lost its duplicated per-kind arms and its post-hoc nudge rewrap; `theme::graph::ELBOW_HEIGHT` is deleted (always exactly `2 * ELBOW_CURVE_SIZE`, referenced only by tests).

Six new tests, three superseded ones removed. The painted-column test was confirmed non-vacuous by reproducing the old nudge position and watching only it fail. All four gates clean: 1119 app + 44 lsp-core + 14 pty-core + 127 wt-core, 0 failed.

**Honest limits.** The measurements are solid. What I could *not* fully derive is the exact snapping arithmetic — my analytical model of `round_to_device_pixel` over these bounds did not reproduce the observed columns, so the parity behaviour here is characterised empirically, not predicted. Fix 2 is deliberately independent of that. What may remain is a **1px seam** where the elbow's stroke hands off to the lane's own rect at fractional scale — a border inset from a rounded box and a filled rect are still two different elements. Killing that class of seam entirely would mean drawing the lane canvas as one path primitive instead of per-element divs, which is a redesign rather than a fix; say the word if you want that. And as always, I have not seen any of this rendered.

If something is still off, the single most useful thing you can send is another screenshot **at the same 1.25x zoom** — that keeps the measurements comparable.

ColinEspinas:
Pushed one more small fix: `Mod+Shift+G` was bound to `root::NewGitGraph` in `default_key_bindings`, but nothing in `AdeApp::render` registered an `.on_action` handler for it, so the keystroke silently no-opped. The command palette's own "Open git graph" row masked this the whole time since it calls `open_git_graph` directly instead of going through the action. Registered the missing handler and added an end-to-end keystroke regression test.

All four gates clean (fmt/build/clippy/test), 1120 app + 44 lsp-core + 14 pty-core + 127 wt-core tests passing. One pre-existing, timing-sensitive flake in `code_surface::diff_view` (unrelated file) showed up under full-suite load and passed cleanly in isolation - documented in BUILD-LOG.md, not touched by this change.

ColinEspinas:
Pushed two more elbow-rendering fixes, both confirmed working live by @ColinEspinas:

- `be4b72f` — the elbow bridge between entry/exit curves was floored at `2 * ELBOW_RADIUS`, which for adjacent-lane elbows (e.g. merging back exactly one row below, or a branch's first diverging elbow) ran the fill past the far curve's own arc, producing a visible notch/stub. Also clamped every arc to stay fully inside its own row and clip the lane canvas per row, closing a GPUI default-`Overflow::Visible` gap where a hovered row's opaque background could paint over a neighbouring row's elbow spillover, making lines appear to vanish on hover.
- `0d907a7` — BUILD-LOG entry documenting root cause + fix for both.

One disclosed, non-blocking limitation carried over from the prior round: a residual ~1px seam at fractional device-pixel scale where an elbow's border-radius stroke hands off to the lane's own filled rect — eliminating that fully would mean drawing the whole lane canvas as one path primitive instead of per-element divs, a redesign rather than a fix. Noted in BUILD-LOG for future reference.

ColinEspinas:
Closing and reopening against the correct current base — this branch has been rebased through the entire Revision R12 stack since this was opened (repo-model, rail-rewrite, Session->Agent rename, worktree-state-safety, title-bar-chips, rail-header-fix, tabs-changes/staging) plus its own §6 fixes (column header, removed session column). Same branch, same content, see the replacement PR.

PR #59, opened 2026-08-01
feat(app): heuristic file-authorship + activity-text engine (Revision R12)

ColinEspinas:
## Summary

Adds `crate::rail::authorship`, a self-contained, GPUI-free module implementing the v1 heuristic approximation for Revision R12's `by: ['s1','s9']` change-row attribution and the rail's live "what is it doing" activity text (`design_handoff_jerry_ade/revision 3/REVISION-2026-07-31.md`, §2.3/§4). Real per-CLI structured tool-call integration is explicitly out of scope for v1 and documented in the module as the upgrade path.

- `attribute_authorship(files, sessions) -> HashMap<PathBuf, Vec<SessionId>>` correlates each changed file's mtime against every session's last-output time, symmetric within a 12s `CORRELATION_WINDOW`. Supports multi-session attribution for the same file. A file with no plausible author is simply absent from the map, never an empty `Vec`.
- `extract_activity(recent_text) -> Option<String>` pattern-matches a gerund verb + filename ("writing auth.rs") or a progress fraction ("bench 3 of 5", "148/312") from a session's recent terminal text, preferring whichever match is most recent. Returns `None` when nothing recognizable is found.
- Both take plain `Duration`/`String` inputs rather than `Instant`/`SystemTime`, so the pure logic is unit-testable without real clocks or a live pty.

Plumbing in `crate::terminal::pane::TerminalPane`:
- A bounded 8KiB tap (`recent_output`) on the pty output already being drained once per poll tick, exposed via `recent_output_text()`.
- `time_since_last_output()`, reusing the existing `activity_at` tracking but not gated on `is_running()`, so a session that just finished writing a file and exited can still be credited for it.

**Not wired into rail state/render yet** - that's a later phase, and depends on the parallel data-model task's `Authorship`/activity shapes landing on `Status`/`SessionRow` first. This module's public API is the intended call surface for both that task and the eventual render wiring. Nothing in this PR claims to show authorship in the UI, and nothing does.

This branch was rebased onto current `master` (through the issue-16 movable-tabs merge, `088f0b0`) - a clean, conflict-free replay, since this module doesn't overlap anything the tab-strip work touched.

## Testing

From this project's usual Linux sandbox (`export LIBRARY_PATH=/tmp/x11-deps/prefix/usr/lib/x86_64-linux-gnu`):

- `cargo fmt --all -- --check` - clean
- `cargo build --workspace` - clean
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo test --workspace --lib -- --test-threads=1` - **1252 passed, 0 failed** across all four crates (`app` 1087, `lsp-core` 44, `pty-core` 14, `wt-core` 107) on the final clean run

One earlier full run hit the known pre-existing `code_surface::diff_view::diff_render_tests` flake (a documented, unrelated timing race - `diff_view.rs` untouched by this change); confirmed it passes cleanly in isolation, and the subsequent full-suite re-run was clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/BUILD-LOG.md
+++ b/BUILD-LOG.md
@@ -3930,3 +3930,55 @@ diff_render_tests` flake - a different test in that module than the two previous
 this repo's own notes (`repeated_refreshes_of_the_same_open_diff_reuse_the_cached_highlighting`
 this time), same timing-sensitive family, `diff_view.rs` untouched anywhere by this change -
 confirmed it passes cleanly in isolation, and the subsequent full-suite re-run above was clean.
+
+## Heuristic file-authorship + activity-text engine (Revision R12)
+
+Adds `crate::rail::authorship`, a self-contained, GPUI-free module implementing the v1 heuristic
+approximation for Revision R12's `by: ['s1','s9']` change-row attribution and the rail's live
+"what is it doing" activity text (`design_handoff_jerry_ade/revision 3/REVISION-2026-07-31.md`,
+§2.3/§4). Real per-CLI structured tool-call integration is explicitly out of scope for v1 and
+documented in the module as the upgrade path - this is a correlation heuristic, not a source of
+ground truth, and is described as such rather than dressed up as more than it is.
+
+- `attribute_authorship(files, sessions) -> HashMap<PathBuf, Vec<SessionId>>` correlates each
+  changed file's mtime against every session's last-output time, symmetric within a 12s
+  `CORRELATION_WINDOW` (the reasoning for that window is documented on the constant itself).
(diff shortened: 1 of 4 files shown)


ColinEspinas:
Holding off on this for now — decided the mtime-correlation heuristic is too unreliable to surface as agent attribution in the Changes panel (risk of confidently showing the wrong session for a file). The diffstat (+/-) itself is real git data and unaffected by this. Branch stays around in case we revisit with real per-CLI structured tool-call data instead.

PR #119, opened 2026-08-02
fix(app): terminal focus and rail idle-collapse bugs when switching terminals from the sidebar

### LINKED ISSUE #112: Bug: switching between terminals in the rail leaves keyboard focus behind, making them appear to merge
When several terminals/agents are open in the same worktree and you switch between them from the rail (or the tab strip), after a few switches the terminals appear to "merge" into one: typing seems to go nowhere, or keystrokes end up triggering app-level shortcuts (e.g. closing a tab) instead of being sent to the terminal you just clicked.

Root cause: `AdeApp::select_agent` (`crates/app/src/work_surface/render.rs:168-216`) only calls `window.focus(...)` when deactivating an open file tab, or when the clicked agent belongs to a different worktree (via `select_worktree`). When switching between two terminals that are in the *same* worktree with no file tab open — the common case for a rail terminal switch — the function falls through to `cx.notify()` without moving focus.

Since `render_center_pane` only mounts the newly active agent's `TerminalPane`, the previously-focused pane's `FocusHandle` is no longer in the rendered frame. GPUI's `focus_node_id_in_rendered_frame` then falls back to `dispatch_tree.root_node_id()`, so subsequent keystrokes dispatch at the window root, outside `TerminalPane`'s `"terminal"` key context. This means:
- Typed input into the "active" terminal can silently go nowhere.
- Global keybindings deliberately scoped to `Some("!terminal")` (e.g. Ctrl+W → CloseFocusedTab, `crates/app/src/lib.rs:511`) become live again, so ordinary keys can trigger app-level actions instead of PTY input.

This exact bug class (an active-agent switch forgetting to move `Window::focus`) was already found and fixed for `Agents::spawn`/`close`/`archive_agent` (`crates/app/src/root/focus.rs:556-650`), but no equivalent fix exists for `select_agent`'s same-worktree switch path — which is exactly what clicking a terminal in the rail does most of the time.

Fix: `select_agent` should always move `window.focus` to the newly active terminal pane's focus handle when switching, not just on cross-worktree switches.

lucasboinet:
Closes #112

## 1. Terminal focus never moved on a same-worktree rail switch

`select_agent` never called `Window::focus` when switching between two open terminals in the same worktree with no file tab active — the ordinary case for a rail/tab-strip click. `Window::focus` stayed pointed at the previously active terminal's now-unmounted `TerminalPane` handle; once GPUI's `focus_node_id_in_rendered_frame` couldn't find that handle in the rendered frame, dispatch fell back to the window root, outside `TerminalPane`'s `"terminal"` key context — so typed input into the terminal that visibly looked selected silently went nowhere, or fell through to normally-suppressed global bindings instead.

Fix reuses this codebase's existing `focus_newly_spawned_agent` focus-restore helper for exactly the missing same-worktree/no-file-tab branch, without touching the two branches that already worked correctly.

## 2. Live follow-up: the rail auto-collapsed the worktree you were actively using

After testing the first fix, a follow-up report described a related but distinct symptom: switching between two terminals in the rail eventually "closed" both, leaving one row labeled with the branch (e.g. `main`) — but the tab strip still listed both terminals the whole time, so nothing was actually closing.

Root cause: `worktree_is_expanded` collapses a worktree's rail row to a single summary line whenever its most urgent agent goes `Idle` (a real, spec'd default — §2.2), with no exemption for the worktree currently selected. A shell reads as idle once it's been quiet a short while, an ordinary occurrence — so the row you were actively looking at could silently collapse mid-use, replacing both terminal rows with the summary line. Purely a rail rendering/collapse-state bug — the real `Agents` list was never touched.

Fixed by exempting the currently selected worktree (`active_agent_cwd() == row.path`) from the idle-collapse default. An explicit caret click still overrides the exemption either way.

## Testing

- `window_focus_follows_a_same_worktree_terminal_switch` (`work_surface::render::tab_scoping_tests`)
- `the_selected_worktree_never_idle_collapses_by_default` (`rail::render::rail_row_tests`) — also proves the caret override still wins
- `worktree_is_expanded_defaults_to_the_real_idle_rooted_rule` adjusted to select a different worktree than the one it inspects, since the fix above would otherwise make it vacuous
- `cargo build --workspace` ✅
- `cargo clippy --workspace --all-targets -- -D warnings` ✅ (0 warnings)
- `cargo fmt --all -- --check` ✅
- `cargo test --workspace`: 1388 passed, 10 failed — all 10 pre-existing and environment-specific (LSP wiring tests needing network/language-server readiness in this sandbox, one GPU-paint check), none touching `work_surface::render`, `rail::render`, or agent-focus logic

### DIFF
--- a/BUILD-LOG.md
+++ b/BUILD-LOG.md
@@ -7154,3 +7154,43 @@ failed** - all 10 failures are pre-existing and environment-specific (LSP wiring
 network access for `npm install typescript@5`, `rust-analyzer`/`pyright` readiness timeouts in
 this sandbox, and one GPU-paint check), none in `work_surface::render` or touching agent-focus
 logic; the new test and every other `work_surface::render` test pass.
+
+## Follow-up on the same live report: the rail was auto-collapsing the worktree the user was actively looking at (GitHub issue #112)
+
+After the focus fix above, a live follow-up report from the same user described a related but
+distinct symptom on the fixed build: switching between two open terminals in the rail eventually
+"closed" both, leaving a single row labeled with the worktree's branch (e.g. `main`) - but the tab
+strip still listed both terminals the whole time, so nothing had actually closed.
+
+Root cause, unrelated to the focus bug: `Self::worktree_is_expanded`
+(`crates/app/src/rail/render.rs`) collapses a worktree's rail row to a single summary line whenever
+its most urgent agent's status is `Status::Idle`, absent an explicit per-worktree override from the
+collapse caret (§2.2: "Worktrees whose most urgent agent is idle start collapsed" - real, spec'd,
+covered by `worktree_is_expanded_defaults_to_the_real_idle_rooted_rule`). A shell agent reads as
+`Idle` once it's gone quiet past `status::RUN_RECENT_OUTPUT_WINDOW` - an ordinary, real occurrence
+(a shell sitting at its prompt between commands), not a fault. The gap: the *currently selected*
+worktree - the one whose terminals the user is actively switching between - was never exempted
+from this default, so its row could silently collapse mid-use, replacing both visible terminal rows
+with the one-line summary the report described. Purely a rail rendering/collapse-state issue; the
+real `Agents` list (and so the tab strip) was never touched, which is exactly why both terminals
+stayed listed there throughout.
+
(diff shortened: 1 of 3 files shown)


lucasboinet:
ca marche pas

lucasboinet:
ca marche

PR #136, opened 2026-08-03
feat(app): predict the next line's indentation on Enter

### LINKED ISSUE #121: Code editor should predict next line indentation
When a user adds a new line the editor should predict where the indentation will be.

lucasboinet:
Closes #121

When a user adds a new line the editor should predict where the indentation will be.

## What changed

`handle_editor_enter_action` previously inserted a bare `"\n"` on `Enter` with no auto-indent. Now it carries the real leading whitespace of the line each cursor is leaving over to the new line (read from the real buffer content, never a hardcoded guess), and adds one more indent unit when that line — after trimming real trailing whitespace — ends with an opening bracket (`{`, `(`, `[`). The indent unit itself comes from `resolved_indent_settings_for_target()`/`indent::indent_unit`, the exact same tabs-vs-spaces/width/`.editorconfig` resolution `Tab` already uses, so `Enter` never disagrees with `Tab` about what "one indent level" means. Multi-cursor aware: each cursor computes its own indentation from its own line.

## Testing

- Three end-to-end tests (`code_surface::editing::editing_tests`) driven through the real bound `EditorEnter` keystroke, covering carry-over, a column-0 no-op, and the opening-bracket case.
- Five `EditBuffer`-level tests (`code_surface::edit_buffer::tests`) covering the same cases plus multi-cursor independence and confirming a *closing* bracket doesn't spuriously add extra indent.
- `cargo build --workspace` ✅
- `cargo clippy --workspace --all-targets -- -D warnings` ✅ (0 warnings)
- `cargo fmt --all -- --check` ✅
- `cargo test --workspace -p app code_surface::`: 352 passed, 4 failed — all 4 pre-existing LSP-server-dependent failures unrelated to this change; all 8 new tests pass.

### DIFF
--- a/BUILD-LOG.md
+++ b/BUILD-LOG.md
@@ -7574,3 +7574,42 @@ indent.
 warnings`, `cargo fmt --all -- --check` - all clean. `cargo test --workspace -p app
 code_surface::`: 348 passed, 5 failed - all 5 pre-existing and environment-specific (LSP/GPU-paint
 tests unrelated to this change, already documented elsewhere in this log); all 8 new tests pass.
+
+## Enter's auto-indent gave Python no extra indent at all (PR #136 review, Colin Espinas)
+
+Colin's review of the auto-indent PR above pointed out a real gap: `indent::ends_with_opener`
+only recognizes `{`/`(`/`[`, so a Python block header (`if x:`, `def f():`, `else:`, `for i in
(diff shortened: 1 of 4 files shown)


ColinEspinas:
Missing indent when no opening character is there like when we use ":" in python ? This should probably be per language ? Idk where we could get the data for each language but this is not complete right now.

Issue #183, opened 2026-08-05
Capture taxonomy: several real captures collapse into coarser buckets (CSS selectors, punctuation.special, label, string.special)


ColinEspinas:
Found by the coverage audit done for the syntax theme redesign (Phase 0). The redesign fixed the one genuinely *unmatched* capture case and added definition-site rules; this issue tracks the remaining, purely-classification gaps it deliberately did not chase.

## Context

`HIGHLIGHT_NAMES` resolves captures by tree-sitter's subset-of-dot-parts rule, so nothing falls through to unstyled text. But several captures resolve only to a much coarser bucket, which means real grammar-level meaning is discarded before the theme ever sees it.

Verified against each grammar's own bundled `queries/highlights.scm` under `~/.cargo/registry/src/`.

| capture | emitted by | currently resolves to | meaning lost |
|---|---|---|---|
| `punctuation.special` | Markdown, Python, JS/TS, YAML | `Operator` | Markdown ATX `#` and list bullets, `${`/`}` interpolation delimiters, YAML `---`/`&`/`*` all become "operator" |
| `label` | Rust, C, YAML | `Variable` | Rust lifetimes, C goto targets and YAML anchors are three unrelated concepts in one bucket |
| `string.special` | JS/TS (regex), TOML (datetime), CSS (colour value) | `String` | a regex literal reads as a string |
| `function.builtin` | Python, JS/TS, Go | `Function` | `len` / `print` / `panic` / `make` |
| `function.macro` | Rust | `Function` | `println!` reads as an ordinary call |
| `tag.error` | HTML | `Tag` | a mismatched close tag reads as normal |
| `constructor` | Rust, Python, JS/TS | `Type` | a `^[A-Z]` heuristic, not a real constructor |

CSS is the lossiest of all: `class_name`, `id_name`, `property_name` and `feature_name` all land in `Property`, and pseudo-selectors in `Attribute`, so a stylesheet renders far flatter than the grammar actually knows how to describe.

Separately: **no** bundled query in any of the twelve grammars distinguishes a type *declaration* site from a type *use*. There is no `@type.definition` anywhere. The redesign added definition-site rules for functions (`@function.definition`); the equivalent for types would need per-language rules in the same way - `(struct_item name: (type_identifier))` / `(class_definition name:)` / `(type_spec name:)` and so on. Sketched but not verified by execution.

## Why it was deferred

[...]

Issue #182, opened 2026-08-05
Bracket ring: one unmatched opener shifts the depth ring for the whole rest of the file


ColinEspinas:
Found while auditing the rainbow-bracket feature for the syntax theme redesign (Phase 0). Split out deliberately: it is a matching/UX problem in the engine, not a colour problem, and fixing it properly is not a palette change.

## What happens

`colorize_bracket_pairs` keeps an opener on its stack until a matching closer arrives. An opener that never gets one stays on the stack for the rest of the file, so every pair *after* it is nested one deeper than it looks and the whole ring is offset from that point on.

Reproduced through the real pipeline (`.` = not a bracket token, `P` = plain/unmatched, digits = ring slot):

```
fn a() { let x = ([)]; }     ....00.P.........P2P2..P
fn b() { let y = ( ;         ....22.P.........P..
}                            P
fn c() { ok(); }             ....44.4...55..4
```

`fn c` is perfectly well-formed code and renders at depths 4/5. The `{` opening `fn a`'s body renders plain, because its `}` got consumed as a shape mismatch against the still-open `(`.

An unterminated string literal does the same thing:

```
fn a() { let s = "oops;      ....00.P...............
fn b() { c([1]); }           ....11.1..23.32..1
```

## Why it matters more than it looks

This is the normal state of a file you are actively typing into. The moment you type `(`, that `(` is genuinely unmatched, so every bracket below the cursor changes colour until you type the closer. The current behaviour is honest but visually noisy in exactly the situation the feature is meant to help with.

`colorize_bracket_pairs`' own docs already state the consequence ("the offset resolves itself the moment the source balances again"). This issue is to decide whether that is good enough, not to report it as undocumented.

## Possible directions (none investigated in depth)

- Bounded lookahead: if an opener has no closer before EOF, drop it from the stack retroactively rather than letting it hold a level.
- Reset depth at top-level structural boundaries, so damage is contained to one item rather than the rest of the file.
- Pair from the tree-sitter tree directly (siblings under one parent) instead of a stack over token order - this is what the redesign spec originally asked for, and it makes the whole class of problem go away, but it needs each injected layer's tree, which the highlight event stream does not expose.

## Out of scope for

PR that follows the syntax theme redesign - referenced there as explicitly deferred.

PR #184, opened 2026-08-05
Redo the syntax theme: a real colour for every semantic category, in OKLCH

ColinEspinas:
## Summary

A redesigned default syntax theme built in OKLCH, an audit + fix of the rainbow-bracket feature (#173), and — after a round of maintainer feedback on the **built app** — a substantial correction of the palette's own direction.

The headline change is now that **every semantic category carries a real, distinct colour**: ten accent hues, all at one OKLCH lightness (0.732) and one chroma (0.105), differing only in hue. An earlier revision of this branch held keywords and function calls at *exactly* plain foreground; that revision was rejected on sight and is walked back here. Details and evidence below.

Along the way the audit found one real matching bug (brackets pairing across separate markdown code fences), one real legibility bug (comments at 3.03:1), one measurable architectural problem (deriving an OKLCH-authored palette through an HSL transform destroyed contrast in the five generated themes), and **six real tree-sitter classification gaps** where a theme token was unreachable in a language. All fixed here.

Builds on #173 — this branch is rebased onto `master` *after* that merge.

---

## The correction, and why

An earlier revision here set `syntax.keyword`, `syntax.function` and `syntax.function_method` to the **literal same hex** as `syntax.text`, on tonsky's "I don't highlight variables or function calls" argument. So `fn`, `let`, `if`, `for`, `return` and every call rendered as prose.

The maintainer built it, looked at it, and rejected it: *"I don't like the new colors at all. I preferred the old colors but I think they were not used correctly."*

Re-reading the two references properly says the same thing:

- **Solarized is not a minimal palette.** It is 8 base monotones **plus 8 genuinely saturated accents** (`#b58900`, `#cb4b16`, `#dc322f`, `#d33682`, `#6c71c4`, `#268bd2`, `#2aa198`, `#859900`). What is disciplined about it is the *construction* — symmetric perceptual-lightness relationships so light and dark keep an identical contrast structure — not a small hue count.
- **Asenov, Hilliges & Muller, CHI'16** is frequently mis-cited as a caution against colour and is the opposite. It is a controlled 33-participant study across three levels of visual richness in which richer visual differentiation cut time-to-answer on structural comprehension questions by **21–75% with no loss of correctness** — contrary to what the participants expected going in. Its literal recommendation is that tools *"use a wider variety of colors by default"* and *"enable the highlighting of more constructs."* Its one real caution is about **non-colour** noise (icons replacing keywords, background tinting everywhere), not about hue count.

**What is not being repudiated is the methodology.** The OKLCH tier, the enforced contrast floors, the bracket cross-injection fix, the Paper black-text-collapse fix and the definition-site query work are all kept. The definition-site work in particular now buys a real *declaration vs call* distinction rather than a *coloured vs not coloured* one.

---

## The palette

Ten accent hues, all at OKLCH **L 0.732 / C 0.105**, differing only in hue. Contrast is against the real editor background `#0d0f11`.

| hue | hex | H | means | contrast |
|---|---|---|---|---|
| red-rose | `#e28c93` | 15 | parameter bindings | 7.69:1 |
| orange | `#de946b` | 50 | constants, numbers, booleans, `self` | 7.82:1 |
| gold | `#c7a356` | 85 | types, JSX tags, markdown headings | 8.05:1 |
| green | `#98b46a` | 126 | strings | 8.30:1 |
| teal | `#4bbeb1` | 185 | attributes / decorators | 8.50:1 |
| cyan-blue | `#51b7d8` | 222 | property / member access | 8.34:1 |
| blue | `#74ade8` | 250 | function and method **calls**, links, caret | 8.13:1 |
| violet-blue | `#a19fe8` | 285 | function/method **definition sites** | 7.90:1 |
| purple | `#c194d6` | 315 | keywords, markdown emphasis | 7.75:1 |
| rose | `#da8db2` | 350 | ordinary variables | 7.71:1 |

Red at H 25 stays reserved for diagnostics, which are a different channel (an underline and a row tint, never a token foreground).

**The tier is not invented.** `L 0.732 / C 0.105` *is* the OKLCH of this app's own original `syntax.function` blue `#74ade8` — the palette the maintainer says they preferred. Every accent above is that colour rotated in hue only, so `function` reproduces the original value **bit-exactly**, `string` lands ΔE 2 from the original `#9dbb6f` (below a just-noticeable difference), and the largest drift anywhere is ΔE 16 on `keyword`, which stays unmistakably the same purple.

Two hues moved further, for a stated reason: `constant` went 67 → 50, because at one shared lightness it sat 18° from `type`'s gold and the two would have collapsed (Solarized separates its own yellow and orange by about the same amount); and `function.definition` is new at 285, the call blue rotated 35°, so a declaration reads as a *distinguished call*.

Neutral ramp unchanged: strong `L 0.885` → text `L 0.762` → comment_doc `L 0.660` → comment `L 0.600` → punctuation `L 0.560`. Comments went from **3.03 → 4.88:1**; punctuation sits at 4.13:1, below plain text on purpose and above the 3:1 floor on purpose.

---

## Six real classification gaps, five of them new

Giving each role a genuinely distinct hue makes a *misclassification* visible for the first time — the lesson of the first one. Every pattern below was compiled and executed against the real grammar, and every node kind and field checked against that grammar's own `node-types.json`.

| gap | effect before | fix |
|---|---|---|
| **Rust has no blanket `(identifier) @variable`** — python/javascript/go/c all do | `syntax.variable` was **unreachable in this app's primary language**. Two rounds of palette work aimed at it could not move a pixel. Found by pixel-diffing two screenshots and getting **zero** changed pixels | `RUST_VARIABLE_PREFIX` (+ `RUST_ATTRIBUTE_SUPPLEMENT` for the one regression it caused) |
| **TS/JS shorthand properties had no capture at all.** `shorthand_property_identifier(_pattern)` appears once in the bundled query, guarded by an all-caps `@constant` predicate; the blanket `(identifier) @variable` cannot reach them (different node kinds) | Every `const { data, error } = useQuery()` and every `return { id, name }` emitted **no highlight event whatsoever**. By blast radius the largest gap found | `TYPESCRIPT_IDENTIFIER_SUPPLEMENT`, with a `#not-match?` guard so an all-caps shorthand still reaches JS's own `@constant` rule |
| **Rust `@constant` is dead upstream.** `tree-sitter-rust-0.24.2/queries/highlights.scm:10-11` reads `(#match? @constant "^[A-Z][A-Z\d_]+$'")` — a stray apostrophe after the `$`, confirmed byte-for-byte with `od -c`. It can never fire, and the later `@constructor` heuristic claims the identifier | Every Rust `const`/`static` rendered as a **type**. Confirmed by sampling the previously committed screenshot rather than inferred: `WRITE_TIMEOUT` measures `#5ec4c4`, the old `syntax.type` cyan — and this repo's screenshot README described those constants as "amber", which was simply wrong | `RUST_CONSTANT_SUPPLEMENT` |
| **`variable.parameter` unreachable in Python and Go** — neither bundled query contains a single one | The parameter hue existed but no Python or Go file could show it | `PYTHON_PARAMETER_SUPPLEMENT`, `GO_CLASSIFICATION_SUPPLEMENT` |
| **TypeScript misses three parameter shapes** — its rules match a *direct* `identifier` child only | `items.map(x => x + 1)` (how most callbacks are written), rest parameters and destructured parameters all read as ordinary locals | `TYPESCRIPT_IDENTIFIER_SUPPLEMENT` |
| **Go: no `@constant`, and composite-literal keys are not `@property`** | `const MaxRetries = 3` read as a local; struct-literal field names read as locals while the identical Rust/TS constructs read as properties | `GO_CLASSIFICATION_SUPPLEMENT`, keyed on the real `const_spec` node rather than on casing |

Two divergences checked and deliberately left alone, recorded in `THEME.md` §2b: Rust emits no `@number` at all (invisible only because number and constant share a colour), and a Rust macro body parses as opaque tokens so `s.field` inside `println!` cannot reach `@property` by any capture rule.

---

## Bracket ring

Six hues at **L 0.560 / C 0.090** — *exactly* `punctuation_bracket`'s own lightness — at the canonical even split **0 / 60 / 120 / 180 / 240 / 300**.

That is a move from the previous `L 0.700 / C 0.080` at hues hidden between the accents, and the reason is arithmetic: with **ten** semantic hue families there is no longer a set of six gaps wide enough to hide in. Held at the old tier, `bracket_5` measured ΔE 9.9 from `function`. So the ring stopped buying separation from hue and started buying it from lightness — which also says something true: a bracket *is* punctuation, and colouring it should convey nesting depth without promoting it above the code it encloses.

It came out **better**, so the floors were raised rather than relaxed:

| comparison | now | previous ring |
|---|---|---|
| adjacent depths (`n` vs `n+1`) | 20.8 (`Moss`) | 20.7 |
| vs plain text | 24.0 | 16.7 |
| vs the de-emphasized bracket tone | 17.6 | 26.5 |
| **vs any semantic accent, all themes** | **12.6** | 11.7 |
| **vs any semantic accent, Jerry Dark** | **20.8** | 11.3 |

`no_ring_colour_is_confusable_with_a_syntax_accent`'s floor went 8 → 12; `no_ring_colour_impersonates_the_semantic_token_it_borrows_its_hue_from` went 11 → 18.

Unchanged and still explicit: the ring rewrites only spans the grammar itself classified `punctuation.bracket`, so a `{` in a string or comment is invisible to it *by construction*; it never touches diagnostics; generics stay plain because tracking `<`/`>` would make HTML actively wrong; injected regions pair independently; `appearance.bracket_pair_colorization` genuinely disables the pass.

---

## Tests: what was replaced, and by what

Two tests asserted the old direction and could not simply be deleted without leaving the palette's structure unpinned:

- `calls_and_keywords_render_at_plain_foreground_in_every_bundled_theme` → **`keywords_calls_and_identifiers_all_carry_real_colour_in_every_bundled_theme`** (ΔE ≥ 18 from plain text, every accent, every bundled theme). Stated in ΔE against plain text rather than contrast ratio deliberately: a colour can have perfect contrast against the *background* and still be indistinguishable from the foreground beside it, which is exactly the failure that got past every contrast assertion in the file.
- `the_palette_spends_at_most_six_accent_hues_plus_two_identifier_tints` → **`every_accent_hue_family_stays_a_real_hue_apart`** (≥ 25°) plus **`no_syntax_token_invents_a_hue_outside_the_accent_wheel`**. A wider palette is only an improvement if its members stay individually identifiable; ten families is a stated ceiling, not an invitation.

Plus six new classification tests, one per gap above, each verified to fail before its supplement and pass after.

---

## Derivation: HSL → OKLCH, plus a contrast guard

Authoring in OKLCH and deriving through HSL was measurably destructive, not merely inconsistent. Measured before the port: **Ember 24 of 39 syntax tokens below 4.5:1, Slate 22**, some as low as 2.15:1, and `Paper` collapsed `text.selected` / `primary` / `heading` into pure black. Those three are now `#3b3e41` / `#414448` / `#484c4f`, **pinned by a test rather than assumed as a side effect of the port**.

OKLCH alone was not enough. `enforce_syntax_contrast_floors` pushes an offending colour away from the background **in lightness only** — holding hue and chroma — measuring the *quantized* colour since an 8-bit file rounds whatever it is given, and one-directionally so a theme that already reads well is returned untouched.

All five derived themes were regenerated. Measured worst cases across all six bundled themes: accents ≥ **4.52:1** against their own editor background and ≥ **ΔE 20.9** from plain text; ring adjacency ≥ **ΔE 20.8**; definition-vs-call ≥ **ΔE 14.5**.

`SEED_REFERENCE_ACCENT` points back at `syntax.function`'s `#74ade8`, which is a genuinely chromatic blue again.

---

## Screenshots

**Per-window `XGetImage` against the app's own toplevel works fine in this environment** (root-window capture returns black); `BUILD-LOG.md` is corrected accordingly, and that stale claim had cost several sessions real time.

`docs/screenshots/jerry-dark-rust.png` was recaptured at the **same file and same scroll position** as the committed one — `crates/lsp-core/src/proc.rs`, lines 136–170 — and pixel-diffed over the code column. **21,097 pixels changed**:

| from | to | pixels | meaning |
|---|---|---|---|
| `#acb2bc` | `#74ade8` | 315 | plain foreground → **call blue** |
| `#acb2bc` | `#c194d6` | 136 | plain foreground → **keyword purple** |
| `#de99be` | `#da8db2` | 231 | old rose → new rose (locals) |
| `#8bc18c` | `#98b46a` | 176 | old string green → new |
| `#88b4ed` | `#a19fe8` | 129 | old definition blue → violet-blue |
| `#5ec4c4` | `#c7a356` | 85 | type cyan → **type gold** |
| `#d8a76d` | `#4bbeb1` | 74 | attribute amber → **attribute teal** |

The first two rows are the whole point: 451 pixels of text that was literally indistinguishable from prose now carry a real hue. `jerry-dark-markdown.png` was recaptured too. `jerry-dark-rust-doc-comments.png` is deliberately left as-is and marked as predating the palette, because it is the *evidence* for the Rust-constant misclassification above.

Scripted navigation is still the unsolved half: the app takes a repo path as `argv[1]`, not a file, so per-fixture captures are not automatable. The substitute remains `fixture_corpus_tests`, which dumps exactly which bucket every byte lands in and what colour and contrast ratio that resolves to, for all six fixtures.

---

## Deferred, with issues filed

- **#182** — one unmatched opener shifts the ring for the rest of the file.
- **#183** — captures that resolve only to a coarser bucket (CSS selectors, `punctuation.special`, `label`, `string.special`, `function.builtin`), plus the observation that no grammar distinguishes a type *declaration* from a type *use*.

---

## Gates

- `cargo fmt --all -- --check` clean
- `cargo build --workspace` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test --workspace --lib -- --test-threads=1` green

Every behavioural change is backed by a test verified to fail before the fix and pass after.


ColinEspinas:
## Screenshots landed

Correcting the "screenshots" checklist item from partial to real: `docs/screenshots/` now contains actual captures of the running app, taken with per-window `XGetImage` against the app's own X11 toplevel.

**`jerry-dark-rust.png`** is the one that matters — a real Rust file with the restraint palette doing exactly what the PR claims:

- `collect_descendant_pids_finds_a_real_child_process` (the **definition** name) is blue, while every `.arg` / `.stdin` / `.spawn` / `.expect` method call around it, and every local (`child`, `descendants`, `pid`, `sh_pid`), is plain foreground
- `Command`, `Stdio`, `Duration`, `Signal` in cyan
- `"sh"`, `"-c"`, `"sleep 30"` in green; `200` in amber
- `let`, `mut`, `fn`, `use`, `for`, `in` all at plain foreground
- the bracket ring visible on real nesting

**`jerry-dark-rust-doc-comments.png`** is the visual counterpart to raising comments off 3.03:1 — a full screen of `///` doc comments at the brighter `comment_doc` tone, alongside screaming-case constants in amber and `usize`/`i64`/`u32` in cyan.

**`paper-chrome.png`** is the light theme. Stated plainly rather than implied: **no editor tab is open in it**, so it shows chrome only, not syntax colours.

### What is still missing

No per-fixture screenshots, and no Paper-with-code capture. The app takes a repo path as `argv[1]`, not a file path, so opening a *chosen* file needs UI interaction; the captures above came from driving the file tree with mouse events, which works but is not reliable enough to automate. `xdotool type --window` mangles characters.

`fixture_corpus_tests` remains the substitute for the per-fixture case, and I would argue it is the better artefact for reviewing *this* PR specifically: it prints, for all six fixtures, exactly which bucket every byte lands in and exactly what colour and contrast ratio that resolves to.

```
cargo test -p app --lib fixture_corpus -- --nocapture
```

ColinEspinas:
## Course correction: ordinary identifiers now carry real colour

Reversed per direct feedback on the rendered result. `syntax.variable` and `syntax.property` are coloured again; function calls and keywords stay plain, since the complaint was about identifiers reading as an undifferentiated wall, not about calls.

### The bug underneath, which is the actual finding

The first attempt at this reversal changed `syntax.variable` and **nothing happened on screen.**

`tree-sitter-rust` is the only grammar here with **no blanket `(identifier) @variable` rule** — python has one at line 3, javascript at 4, go at 26, c at 1. A plain Rust local therefore classified as unstyled `Text`, which made `syntax.variable` **literally unreachable in this app's primary language**. Two rounds of palette work aimed at that token could not have moved a single pixel of a Rust file, and the "most of the text is just white" report was correct both times while being diagnosed as a colour problem.

No contrast or ΔE calculation could have caught it. What caught it was screenshotting after the change, diffing against the one before, and finding **zero changed pixels** in the code area. The same method then caught the regression the fix introduced — the blanket rule claiming the identifiers inside `#[cfg(all(test, unix))]`, because a leaf beats its ancestor and the bundled query only captures the enclosing `attribute_item`.

### Final colours

| token | hex | L | C | H | contrast |
|---|---|---|---|---|---|
| `syntax.variable` | `#de99be` | 0.760 | 0.095 | 346 | 9.01:1 |
| `syntax.property` | `#68bedf` | 0.760 | 0.095 | 225 | 9.63:1 |
| `syntax.variable_parameter` | `#cc9ed7` | 0.761 | 0.094 | 320 | 8.63:1 *(unchanged — already coloured from the binding-site work)* |

Both sit in the palette's existing register (same L and C as every other accent), and land on essentially the hues from earlier this session — warm rose for locals, cool cyan-blue for member access, which is the split that makes an `a.b.c` chain legible. The bracket ring was re-phased to keep clear of them.

### Verified against the render, not just the maths

`docs/screenshots/jerry-dark-rust.png` is replaced with a fresh capture of the **same file at the same scroll position**. Pixel diff over the code column:

- **3297 pixels changed**, dominated by `#acb2bc → #de99be` (plain foreground → rose), 231 instances
- `#d8a76d → #de99be` (the attribute regression): **65 → 0**
- rose is now the third most common text colour in the code area (231 px) after plain text (503) and comments (375)

Visibly: `child`, `descendants`, `sh_pid`, `pid` read as rose where they were white; `#[cfg(all(test, unix))]` is back to amber; the definition name stays blue, method calls and keywords stay plain, types cyan, strings green.

### Honest cost

The semantic set grew from six hue families to eight, so fourteen colours now share one lightness band where twelve did. The bracket ring's distinctness floors are restated from measurement rather than aspiration (worst ring-vs-accent ΔE 8.5, in `Moss`, whose derivation compresses chroma hardest; 11.3 in Jerry Dark). Recorded in the test comments and THEME.md rather than smoothed over — if a future palette wants that headroom back, the lever is fewer semantic hues, not a louder ring.

THEME.md §2 is rewritten to record this as a **reversal**: the binding-site reasoning was sound, it simply produced a screen this editor's user does not want. A future reader comparing this palette against tonsky's essay deserves to know the difference was deliberate.

Also documented: the editor background is actually `surface.pty` `#0d0f11`, not `surface.center` `#131518` which the contrast guard measures against. The guard is therefore conservative — every real ratio is *better* than the one enforced — but the discrepancy is now stated rather than left silent.

Four new tests, each verified to fail before its fix and pass after.

**Gates:** `fmt --check` clean, `build --workspace` clean, `clippy --workspace --all-targets -D warnings` clean, `cargo test --workspace` 1735/1736 — the one failure is the known `rust-analyzer` spawn flake, which passes in isolation and lives in a module this branch does not touch.

Issue #227, opened 2026-08-11
Agent history: repo → worktree → run sidebar view, run-transcript tabs, outcomes and drift


ColinEspinas:
## Summary

Re-scoped 2026-08-14 to the rev-6 design ("Agent triage control room" bundle): agent history is now fully designed — REVISION-2026-08-13.md (the whole document) as amended by REVISION-2026-08-14.md §4/§6 (History moved from a strip cell into the `⋯` overflow; history keyed repo → worktree → run with a scope toggle). Resume/recover groundwork already landed (PR #250: the rail's inline `HISTORY` section with `Resume`/`Reopen`); this issue upgrades that to the designed surface: **the sidebar indexes runs; a centre tab shows one transcript** (audit I8: "Sidebar indexes; center shows one run").

### Entry points

- [ ] History opens from the sidebar strip's `⋯` overflow (#290/#291) — it is a place you visit, not a permanent cell (STAGE-A-CHANGELOG.md §4t).
- [ ] A `↺ N earlier runs` line under a worktree row switches the sidebar to History for that worktree — **only on worktrees with no live agent** (REVISION-2026-08-13.md §6: under every worktree it was "eight identical rows, no information"). Count via #281.
- [ ] The current inline `HISTORY` section inside the rail is deleted in the same change — per REVISION-2026-08-14.md §7 rule 5, quoted verbatim: "Replacing a control means deleting its old keys in the same edit — a key defined twice is two specifications of one thing, and the reader cannot tell which is real."

### The sidebar view (REVISION-2026-08-13.md §3, amended by 08-14 §6)

- [ ] Keyed **repo → worktree → run**, matching the rail's hierarchy, with an `all` / `this worktree` scope toggle; the active worktree carries the selection edge and opens by default (REVISION-2026-08-14.md §6).
- [ ] Two-line run rows: agent chip · title · duration, then outcome pill · drift dot · drift text · finished-at.
- [ ] Empty worktree: *No agent has run in `<branch>` yet.*
- [ ] Filter row placeholder becomes `filter runs` in this view.

### Outcomes — four values, and no `merged` (REVISION-2026-08-13.md §5)

- [ ] `done` / `interrupted` / `failed` / `abandoned` pills (tokens from #279). **No `merged` outcome** — "merging happens to a branch, not to a run"; whether the branch merged is already on the worktree row. Vocabulary aligned with #280 (`finished` live → `done` past).
- [ ] Outcome and drift are independent axes.

[...]

PR #230, opened 2026-08-11
feat(app): a real per-agent review surface, separate from the git diff

### LINKED ISSUE #225: Separate diffs for git and agents
Right now there is a confusion between agents diff review and git diffs.

We need to separate both concepts.

ColinEspinas:
## Summary

Closes #225. Real report: "Right now there is a confusion between agents diff review and git diffs. We need to separate both concepts" — clarified as genuinely separate concepts, not a filtered view of the same data.

Root cause: `wt_core::diff::diff_against_base` answers exactly one question — how does this worktree differ from the merge-base with the default branch — and every "review" surface in the app was fed from it, which can't honestly answer "what has this specific agent changed." An agent that changed nothing in a worktree whose branch already diverged from `main` was reported "Review ready."

**The distinction shipped here is a different base point, not a filter.** An agent's review diff is `git diff <its own baseline tree>` — parsed by the exact same `WorktreeDiff`/`DiffFile`/`DiffHunk` types and the same `git diff` invocation as a real git diff, just with a different base and a different lifetime. This deliberately supersedes the design doc's "one worktree, one git diff, one answer" rule for review specifically — staging/committing stay worktree-level, unchanged.

- **`wt_core::review`**: `snapshot_worktree_tree` takes a real git tree snapshot via a throwaway shadow index (never touches the real index/working tree/HEAD/stash — verified byte-for-byte in tests), anchored under `refs/jerry/review/<sha256 of the baseline key>` so `git gc` can't collect it. `diff_against_tree`/`changed_paths_against_tree` reuse the existing diff parser verbatim.
- Baselines are captured once at agent spawn, and only ever advance when the user clicks **Mark reviewed** — never automatically.
- **Persisted across restarts** (`review-baselines.toml`, merge-on-save so concurrent instances never clobber each other's entries), keyed by `(worktree, agent kind, real wall-clock spawn timestamp)` since neither `AgentId` nor the existing `Instant` field survives a restart. Deliberately doesn't try to solve full agent resume (#227) — just doesn't destroy data that future work could reconnect to.
- **Gated to single-agent worktrees for now**: real per-agent attribution doesn't exist yet, so the whole Review surface (footer door, tab, "Review ready" status) is unavailable whenever a worktree has more than one open agent, re-checked at every relevant decision point through one `review_available_for` accessor.
- A new `Review` tab per agent, patterned directly on the git graph tab's open/close/focus discipline, reusing the existing diff renderer. The header states in words what it's diffed against and when (`claude · since it started · 09:31`).
- Removed now-fully-dead UI: the unwired `Accept file` button and the `Authorship`/author-chip system (its `record()` had zero production callers — the data was always empty).

## A dedicated adversarial review caught two critical bugs before this went anywhere near a push

1. **Unbounded `git add -A`.** The original snapshot mechanism staged every untracked, non-ignored file with real content — measured against this actual checkout, that's 21,471 files / 19 GB hashed into `.git/objects` on every single agent spawn. Fixed: the untracked set is measured first (cheap `git ls-files -o`), and a snapshot falls back to tracked-only past 128 MiB or 5,000 files, with the resulting `UntrackedCoverage` threaded through every later diff and surfaced in the tab header (`· tracked files only`) rather than silently narrowed.
2. **A wedged centre pane.** `leave_review_tab` had exactly one real caller (its own close button) — switching to any other tab while the Review tab was open left it stuck on screen with focus dangling on an unmounted terminal. Fixed at all five real centre-pane takeover sites.

Plus three more real fixes: single-slot `Task` fields that silently cancelled in-flight baseline captures/ref deletions when agents spawned or closed in quick succession (now keyed per-agent); a blocking double-`fsync` on the UI thread (now on the background executor, matching what the sibling persisted-state files actually do); and ref names that could exceed filesystem `NAME_MAX` for realistic deep worktree paths (now a fixed-length SHA-256 digest).

One incidental find worth flagging: chasing this round's own test flakiness root-caused a previously-unexplained flaky-test class in this repo — `fs.inotify.max_user_instances` exhaustion under heavy concurrent `#[gpui::test]` runs that each arm real file-tree watchers, not an event-timing race as previously assumed.

## Test plan

- [x] `wt-core`: real temp-repo tests for snapshot/diff/ref-anchor/ref-delete, including byte-for-byte `git status`/index/HEAD/stash identity before and after a snapshot, an aggressive real `git gc` survival test, and a test asserting real loose-object counts stay near zero when the untracked-set cap kicks in (not just that the flag says so)
- [x] `app`: real `#[gpui::test]`s driving actual spawn/close/mark-reviewed/tab-open flows — including a regression test that drives the real entry points (`select_agent`, opening a file, opening the graph tab) rather than calling `leave_review_tab` directly, since the bug was precisely that those callers didn't, and a concurrency test that spawns/closes two agents with no parking in between so both operations are genuinely in flight
- [x] `cargo clippy -p app -p wt-core --all-targets -- -D warnings` clean
- [x] `cargo test -p wt-core --lib` — 207/207 passed
- [x] `cargo test -p app --lib` — 2100/2100 passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9500,6 +9500,7 @@ name = "wt-core"
 version = "0.1.0"
 dependencies = [
  "gix",
+ "sha2",
  "tempfile",
  "thiserror 1.0.69",
 ]
--- a/crates/app/src/code_surface/diff_view.rs
+++ b/crates/app/src/code_surface/diff_view.rs
@@ -5,7 +5,6 @@ use super::zoom::zoom_scoped;
 use super::*;
 #[cfg(test)]
 use crate::root::focus::palette_focus_tests;
-use crate::root::widgets::render_action_keycap_row;
 use crate::root::widgets::render_sidebar_message;
 use std::rc::Rc;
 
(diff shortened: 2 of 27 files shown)


ColinEspinas:
Rebased past the conflict from #232/#233 (the `AgentKind`/`ProcessKind` type split that merged to master today).

Resolving it surfaced a real design bug in this PR's own review-baseline capture: it captured a baseline (and made the Review tab eligible) for **every** spawned agent, including plain shells - the exact category error #231/#232 exist to eliminate, one layer up in this newer feature. A solo Shell tab could show a "Review" tab purely because some unrelated worktree change happened to land while it was open.

Fixed by gating `capture_review_baseline`/`mark_reviewed` on `ProcessKind::Agent(_)` - a shell never gets a baseline captured, full stop. Rippled through the review test suite's `sole_agent()` helper, which relied on the app's startup shell as its reviewable subject; it now closes that shell and spawns a real `Claude` agent in its place, so every existing test's assumptions keep holding. Full detail in the merge commit message.

All 12 review tests pass for real (not just compile), full workspace suite is clean (2117 passed; the usual 9 known-flaky LSP/file-tree-watch tests under heavy parallel load, all confirmed passing standalone).

Issue #242, opened 2026-08-11
Git graph interactive rebase: verify the merged implementation against REVISION-2026-08-12 §4, then close


ColinEspinas:
## Summary

Re-scoped 2026-08-14 as part of the rev-6 issue audit. The interactive rebase this issue asked for is **built and merged**: the headless engine (phase A, `wt-core`, commit `1e5b650`) and the graph-pane UI mode (phase B, commit `8df91f8`, plus safety-gap follow-up `0f01cd8`) are on master — banner, plan list with drag-reorder, all six action chips with menus, inline reword fields, pause markers, fold elbows, the Result right panel with its warning stack, and the stopped phase. What remains is a **verification pass against the design's own checklist** (REVISION-2026-08-12.md §4 — the spec this issue was opened from), closing this issue on a clean pass or filing targeted deltas for anything that fails.

### Verification — REVISION-2026-08-12.md §4, item by item

- [ ] Rebase is a mode of the graph pane, entered from the row menu, left by Cancel — matches `Jerry.dc.html`'s `Interactive rebase` strip state.
- [ ] Plan is **oldest-first** with the footer order note (`oldest first · applied top to bottom`) — §1.1 is explicit that this must not be "fixed" to match the graph's newest-first order.
- [ ] Six action chips with the spec's palette (rev-6 tokens, #279); menu hints one line each.
- [ ] Fold rows indent with a 1px elbow pointing up-and-left at their target (same vocabulary as the graph's merge elbows).
- [ ] Dropped rows strike through and dim; folded rows dim.
- [ ] `reword` is a real text input, pre-filled, click-stop-propagating.
- [ ] Supplying a reword message removes that pause from the plan, the mark, and the count — and per §1.6's derivation rule: a reword counts as supplied only when non-empty **and different from the original subject** (the pre-filled value is not an answer).
- [ ] Pause column: outlined = planned, filled = actually stopped, legended once in the footer as `pauses here`.
- [ ] `7 → 4 commits` and the whole Result list **derived by walking the plan** (drop skips, squash/fixup fold into the previous entry, everything else appends) — never authored.
- [ ] Running-agent warning appears above the Start button with a working `Pause now` — §1.7: "the one no other git client needs… stated **before** the button, with the remedy attached."
- [ ] Stopped state is the same pane; conflicts go to the **existing** conflict resolver — no second conflict UI (§1.8).

### Wrap-up

[...]

PR #251, opened 2026-08-12
feat(app): interactive rebase UI - graph pane mode (issue #242, phase B)

ColinEspinas:
## Summary

Phase B of issue #242 — the real graph-pane UI on top of Phase A's headless engine (#249, this PR's base — merge in order: master → #249 → this).

- Interactive rebase as a real mode of the graph pane (not a modal): swaps the commit list for an editable plan list, entered from the row menu's "Interactive rebase from here…" (previously stubbed), left via Cancel.
- Real plan editing: drag-to-reorder, per-row action chip (`pick`/`reword`/`edit`/`squash`/`fixup`/`drop`), and a real inline text input for `reword` rows.
- A live-derived Result panel (`N → M commits`, folded/reworded/stop markers) and a warning stack (running agents, remote-tracking, stop count), each rendered only when it actually applies.
- Every action (Start/Continue/Skip/Abort) calls the real `wt_core::rebase` engine and renders whatever real outcome comes back — no simulated state anywhere.
- A genuinely new capability: real agent pause/resume via `SIGSTOP`/`SIGCONT` (process-group-aware, matching the existing `terminate_process_tree` convention) while a rebase is running, since rebasing rewrites files live agents may be editing.
- Found and closed a real gap in Phase A's own engine along the way: its reword-message queue is fixed at rebase-start time, so a message typed *after* a stop needed a new `git commit --amend`-based path (`wt_core::rebase::amend_head_message`) to reach the right commit.

## Review process

This went through an independent correctness review before push (not just self-report), which found 6 real, reproduced Critical issues — most seriously, switching worktrees while a rebase was stopped mid-flight could cause `amend`/`abort`/`continue` to run against the *wrong* worktree's HEAD, and three real paths could leave an agent permanently `SIGSTOP`ped. All six are fixed, each backed by a regression test verified to fail without the fix and pass with it:

1. Worktree/repo switching now clears rebase mode and resumes any paused agents; every op also double-checks a stored worktree root against the live one as a backstop.
2. `Continue`/`Skip` now share the same in-flight guard `Start`/`Abort` already had (fixes a double-click race that could amend the wrong commit).
3. `Cancel` is guarded against abandoning an in-flight `Start rebase` with no recovery path.
4. `amend_head_message` now requires the expected pre-amend HEAD and refuses on staged changes, closing the actual hole that made #1/#2 silent instead of loud.
5. Every rebase-mode exit path (tab close, worktree switch, plan-load failure) now funnels through one `leave_rebase_mode` that resumes paused agents first; "Pause now" is gated against being clickable mid-operation. The one remaining gap — an app crash/SIGKILL leaving a paused process frozen with no next-launch recovery — is a real, accepted limitation, documented rather than silently left.
6. Pause now signals the full process group (mirroring the existing kill-tree logic), not just the direct child — a live agent's own subprocesses (build steps, tool calls) are frozen too, not left running under it.

## Test plan
- [x] 31 new app-level tests (`rebase_flow_tests`, plus `graph_remote_action_tests` for rebase-onto), 22 `wt_core::rebase` tests (2 new for the amend safety checks), 17 `pty-core` tests including a real `/proc/<pid>/status` assertion that pause/resume genuinely suspend and resume a live process
- [x] `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check` all clean (only the pre-existing, unrelated violation at `crates/app/src/review/render.rs:1503` remains, untouched)
- [x] Independent correctness review (not self-report) covering the two highest-risk areas — process pause/resume and the amend-on-continue path — with all findings fixed and mutation-tested


ColinEspinas:
Merged directly into `master` via local conflict resolution (GitHub's stacked-PR merge API reported a genuine merge conflict against master's current state — the same doc-comment/test-count conflicts already resolved once when building this session's integration branch). Landed as commit [`18343c8`](https://github.com/ColinEspinas/jerry/commit/18343c8) plus a follow-up arity fix in [`b7c2192`](https://github.com/ColinEspinas/jerry/commit/b7c2192). Closing since GitHub can't auto-match this PR's exact commit set to what's now in master.

PR #258, opened 2026-08-12
fix(app,wt-core): four multi-repo dashboard bugs found in live testing

ColinEspinas:
## Summary

Fixes four bugs reported from live testing of the multi-repo dashboard (#256/#257):

1. **Couldn't switch from a worktree to another repo's worktree.** `select_worktree_by_path` only ever searched the focused repo's own worktree list. Now falls through to a real cross-repo search across every added repo, checks the target repo out, and synchronously seeds the worktree list from that repo's already-known data so the click takes effect immediately rather than waiting on a fresh background fetch.

2. **Agents silently vanished from the rail** (e.g. after `jerry .` or opening a repo through a symlink). Repo paths were never canonicalized, so a repo opened via a relative or symlinked path never exactly-matched git's own fully-resolved worktree paths - a spawned agent's cwd matched no worktree row and was dropped with no error. Fixed with a single `canonical_repo_path` normalization point applied where a repo path enters the app.

3. **`failed to compute diff: ... fatal: unable to write new index file`**, including a real embedded-git-repo case (`vendor/zed`). The shadow index used for diffing untracked files lived in the OS temp directory; it's now created inside the worktree's own `.git` directory instead, removing a cross-filesystem/quota/cleanup-daemon hazard from the mechanism. Note: I could not reproduce the exact failure locally (tested cross-filesystem/WSL scenarios directly, all succeeded), so this is a real, independently-justified fix, not a confirmed root-cause cure - flagging that explicitly rather than overclaiming.
4. **The rail repo header's `+` button** incorrectly opened the tab-strip's add-agent-tab menu, spawning tabs with no worktree context. Its real intended purpose ("add a worktree to this repo") has no design/UI yet ([standing decision]), so it's now made genuinely inert (disabled visual treatment, no click handler) rather than repurposed to the wrong thing.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] New regression tests per bug, each independently verified to fail without its fix (revert-then-restore) and pass with it
- [x] `cargo test --workspace`: 2291 passed, 12 failed - all 12 in the known, pre-existing `file_tree_watch`/`worktree_watch` inotify-contention flaky class, not new regressions

### DIFF
--- a/crates/app/src/rail/render.rs
+++ b/crates/app/src/rail/render.rs
@@ -914,9 +914,9 @@ impl AdeApp {
     pub(in crate::rail) fn render_rail_list(&self, cx: &mut Context<Self>) -> gpui::AnyElement {
         let groups = self.build_repo_groups(cx);
 
-        // GitHub issue #113: a repo with zero open worktrees is still a real, clickable rail
-        // affordance now (`Self::render_repo_group` renders and wires up every group's header
-        // regardless of `rows`/`all_rows`), so the only case left with genuinely nothing to show
+        // GitHub issue #113: a repo with zero open worktrees still renders its own group
+        // (`Self::render_repo_group` paints every group's header regardless of
+        // `rows`/`all_rows`), so the only case left with genuinely nothing to show
         // is no repo at all - defensive rather than reachable through any real UI path today,
         // since `Self::render_rail` (this function's only caller) is itself only ever rendered
         // once `Self::focused_repo` is `Some`, which requires at least one entry in `Self::repos`.
@@ -953,18 +953,22 @@ impl AdeApp {
     /// waiting` when non-zero, and a per-repo `+`), then either every worktree row already
     /// ranked most-urgent-first by [`rail::group_worktrees_by_repo`], or - GitHub issue #113 - a
     /// real inline message when this repo has none to show, rather than the header (and the repo
-    /// itself) simply disappearing from the rail. Every repo's header renders and is clickable
-    /// regardless of `rows`/`all_rows`: [`Self::checkout_repo_from_rail`] is the same real
-    /// "focus/load a different repo" flow `Self::open_repo_in_current_window` (Open Folder…)
-    /// already uses, so a repo with zero open worktrees is a real, reachable "focused, nothing
-    /// open yet" state - not a dead end.
+    /// itself) simply disappearing from the rail. Every repo's header renders regardless of
+    /// `rows`/`all_rows`, but the header is deliberately **not clickable** - no `on_click`, no
+    /// cursor/hover affordance. That is an explicit product decision, made after two subtler
+    /// header-click behaviors were both rejected in review: in the rail, only worktree rows and
+    /// agent rows are click targets, and only worktrees have tabs; a repo header is a plain
+    /// group label. Switching to a different repo is done by clicking any worktree row under its
+    /// group - [`crate::root::AdeApp::select_worktree_by_path`]'s cross-repo fallback runs the
+    /// entire real repo switch ([`Self::checkout_repo_from_rail`]) itself, so the header never
+    /// needs a click handler for repo switching to work.
     ///
     /// The header's `N wt` and (via [`rail::RepoGroup::waiting_count`]) `N worktrees waiting`
     /// are read from `group.all_rows`, **not** `group.rows` - see [`Self::build_repo_groups`]'s
     /// docs for why: this repo's real, complete worktree list, unaffected by the rail's filter
     /// query or by which repo is currently focused. Only the rows actually rendered below the
     /// header (`group.rows`) may be narrower - and only that narrower list, never the header
-    /// click target or the `+`, is affected by an empty vs. filtered-away distinction (see the
+    /// or the `+`, is affected by an empty vs. filtered-away distinction (see the
     /// inline message below, which does distinguish the two for its own wording).
     ///
     /// `group.rows_loaded` gates the `N wt` count itself: `false` (a repo whose own first real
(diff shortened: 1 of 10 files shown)


ColinEspinas:
Three more bugs found in live testing, fixed in the latest commit:

1. **Repo header still let you open a tab rooted at the repo itself.** Root cause: `checkout_repo_from_rail` left `AdeApp::selected` at `None` until a worktree row was explicitly clicked, so `active_agent_cwd()`'s fallback let the tab strip's "+" spawn a tab with no real worktree behind it. Now lands on the repo's own main worktree as a real selection immediately.

2. **Flicker on the worktree list when clicking a repo** - same root cause: the previous repo's stale rows briefly showed under the newly-focused repo's header before the background fetch replaced them. Fixed by the same synchronous seed as #1.

3. **Diff still failed with `fatal: unable to write new index file`** even after the shadow-index relocation fix, confirmed on this exact branch with a fresh `cargo run --release`, on two unrelated repos - both on native Windows. Root-caused to Windows real-time antivirus transiently holding the freshly-created index lock file at the instant git renames it into place (a well-documented interaction across git GUI tools on Windows). Added a bounded retry scoped narrowly to this one failure text; every other `git add` failure still surfaces immediately, no delay.

All three have real regression tests, each verified to fail without its fix and pass with it.

ColinEspinas:
Course-corrected the previous commit after more live testing (and one honest mistake on my end mid-way):

**Repo header still associated with tabs, take 2.** My first fix auto-selected the repo's main worktree on checkout - which just made the repo header act *as* a worktree click, not fix the underlying issue. My second attempt blocked all spawning while nothing was selected - that broke core functionality (opening a repo and immediately spawning a terminal has never required clicking a worktree row first; sorry for the churn). Reverted both.

The actual fix, per your direction ("repo headers should not be associated to tabs at all"): clicking a repo header is now a pure navigation gesture - it only changes which repo's worktree rows the rail shows. It never selects a worktree and never reactivates an agent into the center pane (new `Agents::clear_active`). Every agent in every repo keeps running and stays visible in the rail exactly as before; only *which* tab (if any) the center pane shows on a header click changed.

**The diff failure, for real this time.** Had a Fable research agent dig into this properly instead of guessing again - antivirus was wrong. The actual bug: `prepare_shadow_index` kept the shadow index file open (as a `NamedTempFile`) for the entire `git add` subprocess. Windows can refuse to rename a lock file over a destination with *any* open handle - a deterministic failure on every git version before 2.48 (Jan 2025), regardless of filesystem or antivirus. Fixed by closing the handle (`into_temp_path()`) before git ever touches the file. The retry from the last commit stays as real defense-in-depth for genuinely external locks, with its comment corrected to stop claiming antivirus as the confirmed cause.

All fixes have real regression tests, each independently verified to fail without the fix and pass with it. Full workspace suite: only the known inotify-contention/timing flaky class, no new regressions.

ColinEspinas:
The repo header is now genuinely non-interactive - no \`on_click\`, no cursor/hover affordance, nothing happens on click at all. My two prior attempts on this same complaint both stopped short of that: one still auto-selected a worktree, the other still re-rooted the sidebar. Per explicit direction, only worktree rows and agent rows are click targets in the rail now; the header is a plain group label.

Repo switching is unaffected - clicking any worktree row under a non-focused repo's group still runs the real switch via \`select_worktree_by_path\`'s existing cross-repo fallback, so there's no loss of functionality, just the header itself no longer being a control.

Real click-through test simulates an actual click on a non-focused repo's header and asserts nothing changes: not focus, not the file tree, not selection, not the active agent, not the agent count. Verified to fail without the fix and pass with it. Full workspace suite: only the known inotify-flaky class, no new regressions.

ColinEspinas:
Fixed one more: "clicking on an agent from another worktree/repo, the tab bar does not appear and only appears when clicking on the worktree."

\`select_agent\` looked up the clicked agent's own worktree only in the *focused* repo's list, but the rail's agent rows show every repo's agents. Clicking an agent from a non-focused repo made it globally active underneath, but nothing switched repos - so the tab strip kept building itself from the *focused* repo's own selection and came up empty. Fixed by delegating to the existing, tested cross-repo checkout (\`select_worktree_by_path\`).

Worth being transparent about: my first version of this fix delegated unconditionally, which introduced a real regression - a test in this same file (\`window_focus_follows_a_same_worktree_terminal_switch\`) caught it immediately, since an agent with no real worktree row anywhere (bare/non-git directory) needs the original plain fallback, not a silent no-op. Corrected before pushing; both the original bug and that regression now have their own passing/failing-without-the-fix regression tests.

ColinEspinas:
Follow-up: #265.

This PR fixed three symptoms of repo-vs-worktree tab attribution (repo-rooted tabs, the header plus-menu with no worktree context, checkout reactivating a stale agent) — each at its own call site. The mechanism that made all three possible survived: `AdeApp::active_agent_cwd` fell back to `focused_repo_path()` whenever `selected` was `None`, quietly making a repo root behave like a worktree it is not.

#265 removes that fallback (`active_agent_cwd` now returns `Option<PathBuf>`) and makes both opening gestures genuinely select a real worktree before spawning the window's initial shell into it — so a focused repo at rest always has a real worktree selected, rather than a `None` limbo that renders a plausible-looking tab.

It also documents two things found live that this PR's own state could still produce: a startup shell attributable to no rail row at all (subdirectory launch), and a rail/tab-strip/centre-pane three-way disagreement after re-checkout. Note this PR's `checking_out_a_repo_from_the_rail_never_selects_a_worktree_on_its_own` carries a doc comment describing the *rejected* auto-select behaviour rather than what its body asserts — corrected in #265, assertions untouched.

The repo header stays entirely inert, and cross-repo agent persistence is unaffected.

PR #260, opened 2026-08-12
fix(app): double-border and selection layout-shift on graph/changes/review rows

ColinEspinas:
## Summary

Reported: selecting a row (git graph, Changes panel, review panel) painted a border on both its left edge and its own bottom separator, and the border shifted the row's content on click.

Root cause: GPUI's `Style::border_color` is a single value shared by every edge of one element (confirmed directly in `gpui`'s own `style.rs`). Rows using both `border_b_1()` (a permanent separator) and a conditional `border_l_2()` (the selection indicator) had the second `.border_color()` call silently overwrite the first - so a selected row's bottom separator flashed the selection colour too, and the border only reserved its 2px of space while selected, shifting content on click.

## Fix

Split the selection edge into a real, separate child element - the same pattern this app's tab strip already uses for its own underline indicator - applied to:
- `crates/app/src/graph_view/render.rs` (git graph rows)
- `crates/app/src/sidebar/render.rs` (Changes panel rows)
- `crates/app/src/review/render.rs` (review panel rows)

A fourth candidate (the rail's agent rows) was investigated and **not** changed: it has only one border edge (no colour collision) and toggled that edge's width between 1px/2px on selection, but a real test proved this never actually shifts the row's content in practice - the row isn't width-constrained the way the other three are. No fix shipped for a bug that couldn't be substantiated.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] New regression test per fixed site, each independently verified to fail without the fix (via revert-then-restore) and pass with it
- [x] `cargo test --workspace`: 2293 passed, 9 failed - all 9 in the known, pre-existing inotify-contention flaky class, not new regressions

### DIFF
--- a/crates/app/src/graph_view/render.rs
+++ b/crates/app/src/graph_view/render.rs
@@ -1715,21 +1715,21 @@ impl AdeApp {
             .items_center()
             .w_full()
             .h(theme::graph::ROW)
-            .border_b_1()
-            .border_color(theme::border::ROW)
             .cursor_pointer()
-            // The left selection edge is a real, separate child - not `border_l_2()` - because
-            // GPUI's `Style::border_color` is one shared value for every edge of a single
-            // element (confirmed directly in `gpui`'s own `style.rs`: `border_color:
-            // Option<Hsla>`, not per-edge). `border_l_2()` on this same div would silently
-            // overwrite the `border_b_1()` separator's colour above with whichever colour
-            // selection picks - toggling *both* edges together, so a selected row would flash a
-            // second border along its own bottom, not just gain one on the left. Splitting them
-            // into two real elements is what actually fixes that, on top of the layout-shift fix
-            // this already had: `left_0().top_0().bottom_0()` reserves the same 2px of space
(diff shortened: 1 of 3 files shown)


ColinEspinas:
Follow-up per direct clarification: these rows should have **no bottom border at all**, not just a fixed/non-colliding color for it. Removed \`border_b_1()\` entirely from graph rows, Changes panel rows, and review panel rows - only the left selection edge remains.

For graph rows this also cleaned up a real, previously-delicate pixel-alignment fix: the row's border used to make its content box 1px shorter than its border box (GPUI is border-box), which needed careful half-pixel-offset handling in \`render_graph_lane_canvas\`. With no border at all, that discrepancy is gone. Full graph_view suite (115 tests, including the whole-pixel-alignment and lane-tiling regression tests) passes unchanged - confirms this doesn't reintroduce the earlier "lines and elbows misaligned" bug those tests guard against.

PR #269, opened 2026-08-12
fix(app): a shell tab keeps its program name when an agent joins the worktree

ColinEspinas:
The shell tab reads `zsh · master` on a bare worktree, then **silently relabels itself to the literal `terminal`** the instant an agent is spawned *beside* it. Nothing about the terminal has changed.

## Root cause

`current_worktree_agent_tab_labels` recomputes every label from scratch on each render, and the `TabChipKind::Term` arm switched on `current_worktree_is_bare()` — **a property of the whole worktree, not of the tab** — handing back a hardcoded `"terminal"` whenever it was false. So opening an agent anywhere in the same worktree retroactively strips the shell tab's identity.

## `"terminal"` was never the design's label

The design's own reference implementation:

```js
{ id: 'terminal', label: prim ? 'zsh' : 'zsh · ' + activeWt.branch, mono: true, kind: 'term' }
```

The label is the **program in both branches**; only the ` · <branch>` suffix drops. `'terminal'` there is the tab's **id** — which is almost certainly where the literal came from. R12 §3 only ever specified the bare case, so the non-bare label had no spec backing.

## The fix

Pass the branch only while bare, reusing `bare_worktree_shell_label`'s own documented `None` fallback:

```rust
work_surface::TabChipKind::Term => work_surface::bare_worktree_shell_label(
    &agent.pane.read(cx).program_label(),
    if is_bare { branch.as_deref() } else { None },
),
```

Bare → `zsh · master`; non-bare → `zsh`. `disambiguate_tab_labels` then yields `zsh #1` / `zsh #2` for two shells in a non-bare worktree, keeping the spec's "never two tabs with the same label" rule.

## Tests

New regression test `a_shell_tab_keeps_its_program_name_when_an_agent_joins_the_worktree`, **confirmed to fail against the old code** (`left: "terminal"`, `right: "zsh"`) rather than passing vacuously. Existing `a_bare_worktrees_shell_tab_label_joins_the_real_program_with_its_branch` untouched and still green. `work_surface` 108/108.

## Heads-up, unrelated to this PR

`master` is currently **red on `cargo clippy --workspace --all-targets -- -D warnings`** on Linux:

```
error: variable does not need to be mutable --> crates/app/src/lib.rs:206:9
```

Note that clippy's suggested fix (removing `mut`) **would break the macOS build** — the `mut` is genuinely required by the `#[cfg(target_os = "macos")] bindings.push(...)` at `lib.rs:609`. It only lints on non-macOS. Fixed separately.

### DIFF
--- a/crates/app/src/work_surface/render.rs
+++ b/crates/app/src/work_surface/render.rs
@@ -864,8 +864,9 @@ impl AdeApp {
 
     /// The tab label each of `agent_ids` (already filtered to the currently selected worktree)
     /// should render, in the same order - real `TerminalPane::program_label` facts (the
-    /// bare-worktree shell gets `work_surface::bare_worktree_shell_label` instead of the generic
-    /// `"terminal"`) run through `work_surface::disambiguate_tab_labels`, computed once so
+    /// shell gets `work_surface::bare_worktree_shell_label`, which adds the ` · <branch>` suffix
+    /// only while the worktree is bare) run through `work_surface::disambiguate_tab_labels`,
+    /// computed once so
     /// [`Self::render_tab_strip`] and this method (also used directly by tests) can never
     /// disagree about what's actually rendered. Any id not found in `Self::agents` is skipped
     /// rather than panicking - `render_tab_strip` re-does the same lookup per id and simply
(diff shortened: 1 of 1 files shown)


ColinEspinas:
Closing in favor of a proper fix: tab labels should come from the terminal's real, live OSC-title signal (`TerminalPane::title()`), not a static `program_label()` + branch-suffix + disambiguation-number scheme. This PR patched the old scheme rather than replacing it. Follow-up incoming.

Issue #289, opened 2026-08-13
Rail refresh: repo-header urgency dots, selection-only edges, hierarchy, prune icon


ColinEspinas:
## Summary

Rev 6's rail corrections (REVISION-2026-08-14.md §4 + STAGE-A-CHANGELOG.md §4k–§4s). The two structural reversals against today's code: the worktree row's 2px left edge becomes **selection-only** (today it carries the most-urgent agent's status colour — the design explicitly reversed that in §4m), and the repo header's amber "N worktrees waiting" **text** becomes **two separate dot+count pairs** — amber for needs-input, red for failed. Plus a set of smaller corrections: hierarchy/spacing, coloured diffstats, a bigger caret, neutral ages, and prune as an icon.

### Repo header (REVISION-2026-08-14.md §4, STAGE-A-CHANGELOG.md §4q/§4s)

- [ ] 24 high, padding `10 12 6`; bare `border-top` rule, **no fill** — "a filled header reads as a container"; the rule sits above the label; the first repo band carries no top rule at all (§4u: the filter row's border already ends the chrome).
- [ ] Label at the app-wide section-label token (the same one the panel section headers use, §4s) — name is **the only shrinkable thing in the row** (ellipsis); both counts never shrink or wrap.
- [ ] `N wt` count (existing behaviour, including the honest em-dash before rows load).
- [ ] **Two separate urgency counts**, right-aligned: amber dot + count (worktrees needing input) and red dot + count (worktrees with a failed agent) — each hidden at zero, each with its own tooltip sentence. A worktree holding both counts **once, as failed**.
- [ ] Rule quoted verbatim (REVISION-2026-08-14.md §7 rule 4): "Two states distinguished anywhere in the app are never summed anywhere in it."
- [ ] The old `waiting_count_label` text path is deleted in the same change (§7 rule 5, quoted verbatim: "Replacing a control means deleting its old keys in the same edit — a key defined twice is two specifications of one thing, and the reader cannot tell which is real.")
- [ ] Header sits closer to its own rows than the rows sit to each other (§4s spacing).

### Worktree rows (STAGE-A-CHANGELOG.md §4m)

- [ ] The 2px left edge means **selection or nothing**: selection-blue on the active row, transparent otherwise. The status-coloured edge, the prunable edge and the bare edge are all deleted (residue of the status channel — "a channel with one meaning has exactly two states, on and off").
- [ ] Agent rows keep their status edge — there the status belongs to the row's object.
[...]

PR #313, opened 2026-08-14
Rail: fix agent/history row indent + filter row padding drift (#289 follow-up)

ColinEspinas:
## Summary

Stacked on PR #311 (issue #289). Two real gaps found by independently comparing the whole rail against `Jerry.dc.html` after #289 landed:

1. **Agent row indent** — the live agent row's status edge was flush at x=0 instead of indented 14px under its worktree (13px indent + 1px connector). An earlier cut folded the indent (padding-left) and the status edge (border-left) onto the same div; GPUI draws a border at a box's own outer edge same as CSS, so that put the edge at the row's own x=0 instead of inset. Restructured to match the mock's real shape: outer padding-left wrapper → 1px connector → content box with its own border-left.

2. **History row, same bug class** — `render_past_agent_row` had the identical structural issue, plus a hardcoded neutral edge color instead of the real per-row status color, plus padding that didn't match the mock's asymmetric `6 10 7 7`.

3. **Filter row padding drift** — 10px horizontal padding where the mock uses 12px (`Jerry.dc.html:109`).

## Test plan
- [x] New real geometry test (`the_agent_row_is_really_indented_under_its_worktree_not_flush_with_it`) measures actual painted bounds, not source
- [x] Personally mutation-tested: reverted the fix, confirmed the exact expected failure, restored
- [x] `cargo build --workspace` / `clippy -D warnings` / `fmt --check` clean
- [x] Full rail suite green (171-177 tests); known inotify-contention watcher flakes confirmed unrelated (system-wide instance exhaustion from concurrent agents at the time, re-confirming once that clears)

### DIFF
--- a/assets/themes/ember.toml
+++ b/assets/themes/ember.toml
@@ -170,10 +170,12 @@ tree_modified = "#8a5c21"  # "M" mark
 # spine/selection edges) already has one, reused directly at the call site rather than
 # duplicated here under a second name.
 [rail]
-repo_header_name   = "#576064"
+repo_header_name   = "#6e777b"  # Repo group header's uppercase name.
 worktree_active_bg = "#151a1c"
 worktree_hover_bg  = "#14181a"  # Worktree row hover background (§2.2: "hover #16191c").
-prunable_edge      = "#252c2f"
+agent_title        = "#747c81"
+repo_ask_count     = "#a66730"  # The repo header's amber urgency **count**
+repo_fail_count    = "#984d60"  # The repo header's red urgency **count**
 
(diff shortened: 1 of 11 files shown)


ColinEspinas:
Superseded by #319 — GitHub auto-closed this PR when its base branch `issue-289-rail-corrections` was deleted on merging #311, and refuses to reopen it because the head branch was force-pushed (rebased onto master) afterwards. #319 is the same branch and the same work, based directly on master, plus a design-fidelity follow-up commit on the row edge.

Issue #287, opened 2026-08-13
Attribution UI: per-agent diff gutter, author chips, and the shared-file ring filter


ColinEspinas:
## Summary

The visible half of rev 6's differentiator (REVISION-2026-08-14.md §1, AUDIT-2026-08-13-competitive-v2.md §3.2): per-line gutter attribution in the diff view, author chips on shared files, and the `⚠` ring as the entry point to a per-author filter. Built on the provenance backend (#284). Today the diff view (`code_surface/diff_view.rs`) has no author channel of any kind, and change rows carry no chips or ring.

### Gutter attribution

- [ ] Each diff line in the Uncommitted/Runs scopes renders a thin gutter bar in its author's agent tint; `you` renders the neutral hand-edit token and is labelled *you — hand edit* (tooltip).
- [ ] Adjusts the existing `DiffRow` line renderer (which already draws a diff-kind accent bar) — the author gutter is an additional channel, not a replacement.
- [ ] Attribution renders only where provenance exists; unattributed lines carry no bar (never guessed).

### Author chips and the `⚠` ring (REVISION-2026-07-31.md §4, STAGE-A-CHANGELOG.md §1)

- [ ] Each Uncommitted row gains a chip group after the filename: one agent chip per author, reusing the existing agent chip component.
- [ ] Chips render **only when the worktree has more than one agent** — "with one agent every chip is identical and carries no information" (REVISION-2026-07-31.md §4).
- [ ] A ring around the chip group when the file has more than one author. **The ring stays amber** — product decision 2026-08-14, overriding STAGE-A-CHANGELOG.md §4g's "flagged for a later call" note: the user's call is that a shared file genuinely needs attention, consistent with the app's amber-means-attention language.
- [ ] The ring is a control, with the design's tooltip: *Two agents edited this file — click to filter the diff by author.*
- [ ] There is no `⚠ N` count on the rail worktree row (STAGE-A-CHANGELOG.md §4g removed it — the signal lives on the file row, where it is actionable).

### The per-author filter

- [ ] Clicking the ring opens the file's diff filtered to one author; other authors' lines dim (the mock uses 0.32 opacity); the scope line appends `· <agent> only` with an `✕` to clear.
- [ ] While a filter is active a single dismissible `<agent> only ✕` indicator shows in the file toolbar — only while active (STAGE-A-CHANGELOG.md §4b) — so a filtered diff can never read as the whole diff.
[...]

Issue #281, opened 2026-08-13
One pluralisation helper for every count in the window


ColinEspinas:
## Summary

Rev 6 mandates one pluralisation helper for every derived count in the window (REVISION-2026-08-13.md §8a, restated in REVISION-2026-08-14.md §6). The app currently has **no** such helper — every counter conjugates inline (`rail/state.rs` `waiting_count_label`, the title-bar chips' `agent`/`agents` + `needs`/`need`, the rail's `{n} file{s}`, status-bar clusters, …), which is exactly the drift the design caught in its own mock ("the two that were written by hand both said `1 files`").

The governing rule, quoted verbatim from REVISION-2026-08-14.md §7:

> 9. Every count goes through the pluralisation helper; never inline a ternary.

### The helper

- [ ] One function, `n(count, singular, plural?)`-shaped (Rust naming at the implementer's discretion), returning `"1 file"` / `"3 files"`; default plural = singular + `s`, explicit plural for irregulars.
- [ ] Lives with the other shared widgets/helpers (`root/widgets.rs` or a sibling), documented with the §7 rule above.
- [ ] Handles the sentence-conjugation cases the title bar needs (`1 agent needs input` / `2 agents need input`) — either via the helper or a thin companion, but through one code path, not per-call-site ternaries.

### The sweep

- [ ] Every existing inline conjugation is converted: rail repo-header waiting label, rail file counts, title-bar agent chips, status-bar clusters (`N agents`, `N wt`, `N servers · M errors`), palette counts, settings counts, prune labels, tab labels — audit by grepping for `"s"`-append and ternary-plural patterns.
- [ ] A short doc note (CLAUDE.md or module doc) records the rule so new counts use it.
- [ ] Singular and plural both verified for at least: 0, 1, 2 on each converted site (the design calls out drift bands and rail links specifically needing "singular and plural both, everywhere", REVISION-2026-08-13.md §4).

### Acceptance

- [ ] No user-visible string anywhere reads `1 files`, `1 agents`, `1 worktrees`, etc. — spot-checked against `Jerry.dc.html`'s `Rail · triage` and `Run · transcript` states.

### Dependencies

None. Every other rev-6 issue routes its new counts through this helper (their bodies assume it exists).

Design sources: `design_handoff_jerry_ade/revision 5/REVISION-2026-08-13.md` §8a, `REVISION-2026-08-14.md` §6/§7 rule 9.

PR #327, opened 2026-08-15
Diff-line review notes: draft/sent, batched, sent to the run's agent (#288)

### LINKED ISSUE #288: Diff-line review notes: draft/sent, batched, sent to the run's agent
## Summary

Rev 6 adds diff-line review notes — audit top-5 #2 (AUDIT-2026-08-13-competitive-v2.md §6): line-anchored comments, **batched**, sent to a **named agent**, and **kept pinned after the revision** so you can verify. The audit's sourcing (Orca's Annotate-AI-Diff, "including the parts they learned the hard way"): *"sending comments one at a time causes the agent to swing back and forth"* — hence one batch, sent from the top of the diff. Nothing like this exists in the app today (no comment model anywhere).

### Notes on lines (STAGE-A-CHANGELOG.md §1, §6.2 rows)

- [ ] Clicking a diff line (bound to `C` per the footer hints) toggles an authored note pinned beneath that line — a card with the selection-blue left edge, the note author chip, and a `draft`/`sent` mark.
- [ ] Notes interleave into the existing `DiffRow` stream (adjust the enum; reuse the diff renderer — no parallel list).
- [ ] Notes are keyed per worktree + path + line and survive scrolling/reopening.

### Batched, from the top (STAGE-A-CHANGELOG.md §1)

- [ ] A notes bar above the hunks: the note count, the design's line verbatim — *one prompt, line-anchored · pinned after the revision* — and a `Send notes to <agent>` button with `mod+enter` keycaps.
- [ ] Sending delivers **one** batched, line-anchored prompt into the target agent's pty — never one message per note.
- [ ] **Target = the run's agent:** `noteTarget` resolves to the file's first author, falling back to the worktree's primary agent — "in a shared worktree the notes go to whoever wrote the lines" (per #287's author data).

### Pinned after send

- [ ] Send flips each card's mark `draft → sent` and the bar to *N notes sent — awaiting revision* with a sent confirmation; **it never clears the notes** — they stay pinned so the revision can be verified against them.
- [ ] Editing after send starts a new draft state for that note.

### Acceptance — the mock's sad path as criteria

- [ ] `Jerry.dc.html`'s `Review · uncommitted` state demonstrates the full loop and both states: a note pins on click, the bar reads `1 note on this file`, sending flips it to `1 note sent — awaiting revision`, the card's mark flips `draft → sent`, and **the note is still pinned** (STAGE-A-CHANGELOG.md §3 verification; STAGE-A-SELFCHECK.md demo data: "review notes in both `draft` and `sent` states").
- [ ] The send shortcut renders as keycaps; every control here carries a tooltip (ride-alongs I10/I11).
- [ ] All counts through the pluralisation helper (#281).

### Dependencies

Depends on: #285 (Runs section context), #287 (author resolution), #284 (provenance), #279, #281. Related: #7 (conversation bookmarking is adjacent but distinct — no overlap in scope).

Design sources: `design_handoff_jerry_ade/revision 5/AUDIT-2026-08-13-competitive-v2.md` §6.2 / top-5 #2, `STAGE-A-CHANGELOG.md` §1/§3, `STAGE-A-SELFCHECK.md`.

ColinEspinas:
Audit top-5 #2 (`AUDIT-2026-08-13-competitive-v2.md` §6.2), built as `STAGE-A-CHANGELOG.md` §1's two `§6.2` rows describe it and verified against §3's own acceptance list. Nothing like this existed in the app before — there was no comment model anywhere, and no way at all to say something back to an agent about a specific line.

## What it does

Orca's `Annotate AI Diff`, including the three parts the audit says they learned the hard way — each structural here rather than a rule someone has to remember:

- **Batch.** *"sending comments one at a time causes the agent to swing back and forth"*. `review_notes::prompt::BatchedPrompt` composes every note on a file into one string, and `TerminalPane::send_prompt` delivers it with exactly one submit byte. It also **refuses** a payload that could split: with bracketed paste off, `paste_payload` turns every `\n` into the Enter byte, so a five-line batch would arrive as five separate messages and look like it worked. `for_delivery` therefore has two forms — multi-line when the target really has `DECSET 2004` on, `·`-joined with no line breaks when it does not — and `send_prompt` checks that invariant rather than trusting it.
- **Send from the top.** A notes bar above the hunks: the count (through #281's pluralisation helper), the design's line verbatim — *one prompt, line-anchored · pinned after the revision* — and `Send notes to <agent>` with real `mod+enter` keycaps behind a real binding (ride-along I10). Every control carries a tooltip (I11).
- **Keep them pinned.** Sending only ever *sets* `ReviewNote::sent`; there is no code path in the folder that removes a note as a consequence of sending one.

Notes interleave into the existing `DiffRow` stream as a new `DiffRow::Note` variant — no parallel list, one diff renderer — and are anchored to a real **file line number** (`NoteAnchor::New`/`Old`) rather than a row index, so a hunk growing above them cannot silently re-anchor every note. Removed lines are annotatable and keep their own old-file anchor. State persists in `review-notes.toml`, keyed worktree → path → anchor, merged under the shared lock.

`draft`/`sent` is derived, not stored: `ReviewNote::sent` holds the exact wording delivered, so editing a sent note goes back to `draft` while the history stays honest about what the agent really was told.

**Target resolution** is §1's `noteTarget`: the file's first author from #287's provenance, resolved to a live agent session in this worktree, falling back to the worktree's primary agent — now one rule, `Agents::primary_for_cwd`, factored out of `activate_for_worktree` so the centre pane and a send target cannot disagree. A shell is never a target.

## Two deliberate departures from the mock, both documented at the code

- **The card is one row tall.** Since #224 the diff is a `gpui::uniform_list`, which lays every slot out at the height measured from item 0, so a free-height wrapping card is not representable; the multi-slot alternative would have to guess a line count before layout knows the pane width, and would vanish the moment its head scrolled off. Same four elements in the same order, text on one line with the whole of it in a tooltip.
- **The selection-blue left edge is a real 2px child**, because GPUI carries one `border_color` per element and the mock's two-colour border has no single-element equivalent.

## Verification

The mock's own acceptance sequence, end to end in a real window against a real spawned agent process: click pins a card → bar reads `1 note on this file` → send → bar reads `1 note sent — awaiting revision` with `✓ sent`, the card's mark flips `draft → sent`, **the note stays pinned** — and the batched prompt really arrives in that agent's real pty, observed through the child's own output rather than by asserting `write_input` was called. Done both as a `#[gpui::test]` and as a live X11 run with screenshots.

The second commit closes eight defects an adversarial review found, including one critical: the note's focus node used to live on the card, so scrolling it out of the virtualized list deleted the focused node mid-sentence and every keystroke went silently nowhere. Also fixed: `v` and `]` (bound under `"diff"`) were swallowing characters typed into a note; a note being typed was lost on quit; `sent` could be claimed for a delivery into a dead agent; a draft was not bound to a checkout. See that commit's message for all eight.

Closes #288

### DIFF
--- a/assets/themes/ember.toml
+++ b/assets/themes/ember.toml
@@ -226,6 +226,29 @@ critical = "#984d60"  # Below 15% remaining.
 track    = "#1c2124"  # The unfilled remainder of the meter
 stale    = "#4e565b"
 
+# Diff-line review notes (GitHub issue #288, STAGE-A-CHANGELOG.md §1's two §6.2 rows and the
+# audit's top-5 #2) - the notes bar above the hunks and the card pinned beneath a line.
+[notes]
+edge             = "#00789d"
+card_bg          = "#12191c"  # The pinned card's fill.
+card_border      = "#1a333f"
+card_fg          = "#8a9194"  # The card's own note text.
+card_placeholder = "#454d51"  # The placeholder shown in an empty, still-being-typed card.
+mark_draft       = "#488199"  # A card's draft mark
+mark_sent        = "#689057"  # A card's sent mark, and the bar's own ✓ sent confirmation.
+dot              = "#518fa8"
+dot_empty        = "#2d3336"  # The same column's ○
+bar_bg           = "#11191d"  # The notes bar's band.
+bar_border       = "#152b35"  # Its 1px bottom rule against the hunks below it.
+bar_label        = "#6599ae"  # The bar's count sentence (1 note on this file).
+bar_meta         = "#454d51"
+bar_error        = "#984d60"
+send_bg          = "#112027"
+send_border      = "#144a5c"  # Its 1px border, and its keycaps' border.
+send_hover_bg    = "#132830"  # Its hover fill.
+send_fg          = "#6599ae"  # Its label.
+send_cap_fg      = "#488199"  # Its ⌘⏎ keycaps' glyphs.
+
 
 # ──────────────────────────────────────────────────────────────────────────────────────────────
 # The code surface
--- a/assets/themes/jerry-dim.toml
+++ b/assets/themes/jerry-dim.toml
@@ -226,6 +226,29 @@ critical = "#b46d61"  # Below 15% remaining.
 track    = "#292c30"  # The unfilled remainder of the meter
 stale    = "#676c72"
 
+# Diff-line review notes (GitHub issue #288, STAGE-A-CHANGELOG.md §1's two §6.2 rows and the
+# audit's top-5 #2) - the notes bar above the hunks and the card pinned beneath a line.
+[notes]
+edge             = "#5e8dc4"
+card_bg          = "#1d2126"  # The pinned card's fill.
+card_border      = "#323f51"
+card_fg          = "#b0b4b9"  # The card's own note text.
+card_placeholder = "#5c6166"  # The placeholder shown in an empty, still-being-typed card.
+mark_draft       = "#7a9abe"  # A card's draft mark
+mark_sent        = "#70b693"  # A card's sent mark, and the bar's own ✓ sent confirmation.
+dot              = "#88abd2"
+dot_empty        = "#3d4146"  # The same column's ○
+bar_bg           = "#1c2127"  # The notes bar's band.
+bar_border       = "#2a3544"  # Its 1px bottom rule against the hunks below it.
+bar_label        = "#99b8d9"  # The bar's count sentence (1 note on this file).
+bar_meta         = "#5c6166"
+bar_error        = "#b46d61"
+send_bg          = "#202933"
+send_border      = "#3c5975"  # Its 1px border, and its keycaps' border.
+send_hover_bg    = "#26323e"  # Its hover fill.
+send_fg          = "#99b8d9"  # Its label.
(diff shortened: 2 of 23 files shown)


ColinEspinas:
Merged `master` in. One conflict worth calling out because it changes an existing binding rather than only this branch's own:

#329's `space` → `ToggleChangeStaged` landed under `Some("diff && !file-editor")`, the same predicate `v` (`ToggleChangeSeen`) and `]` (`NextChangedFile`) carried before this branch narrowed them. All three now carry `&& !text-input`, for one reason: this issue's pinned review-note card is a real `"text-input"` node **inside** `code_surface::render`'s `"diff"` node, and `"file-editor"` does not cover it. Without the conjunct, typing into a note marks the file seen on every `v`, jumps to another file on `]`, and stages/unstages on every space — GPUI dispatches a matching `KeyBinding` before any `on_key_down` listener, so it is not a listener-ordering question.

It is a no-op for every context stack that existed before this branch: the only `"text-input"` node under `"diff"` was the File editor, which already carries `"file-editor"` on the same node.

`plain_letters_bound_over_the_diff_are_typed_into_a_note_not_swallowed` types a string containing a `v`, a `]` and real spaces, and asserts all of it landed in the note, nothing was marked seen, nothing was staged, and the open file did not change.

Issue #414, opened 2026-08-16
Replace design_handoff_jerry_ade/ with a real, evolving docs/design/ set


LucasPcq:
Jerry's UI was built from a one-shot design handoff bundle: `design_handoff_jerry_ade/` — 9 tracked
files (536 KB) holding two copies of the same artefact: a 4,266-line interactive HTML mockup
(`Jerry.dc.html`), a 3,822-line `support.js`, a 415/420-line README written as a build brief, and a
`tokens.rs` of transcribed hex constants.

`CONTRIBUTING.md` already describes it honestly — *"a one-time handoff artifact, not a living
spec."* It has become the de-facto design authority anyway: `.claude/skills/{plan,verify}`,
`docs/development-workflow.md` and `.github/ISSUE_TEMPLATE/feature_request.md` all send readers to
it as "the authoritative reference", and **363 doc comments across 81 files in `crates/`** cite it.

## Three problems

**1. ~111 of those 363 citations already point at directories that were never committed.**
`revision 2/`, `revision 3/`, `revision 5/`, `STAGE-A-CHANGELOG.md`, `REVISION-2026-08-13.md`,
`REVISION-2026-08-14.md`, `AUDIT-2026-08-13-competitive-v2.md` — none of these exist in the repo. An
outside contributor reading `theme.rs` or `sidebar/render.rs` today hits an unresolvable path
roughly every other screen.

**2. The citations are exactly what `CLAUDE.md` forbids in source comments** — "design history,
issue archaeology, revision IDs". This is the single largest class of comment-rule violation in the
codebase.

**3. A frozen mockup cannot absorb a design change.** Every evolution since has been recorded as
another undelivered `revision N/` folder, or not at all. #282 (phosphor icon pack) silently
superseded the handoff's "every icon is composed from rects and text glyphs" rule with nothing to
record it in.

## Outcome

`design_handoff_jerry_ade/` deleted, `docs/design/` in its place: a structured design documentation
set a new open-source contributor can read, and that a UI PR **updates** instead of contradicts.
Prep work ahead of a broader repository cleanup.

## Decisions taken

- **`crates/app/src/theme.rs` stays the source of truth for values.** Its ~270 tokens already carry
  doc comments and already feed `docs/themes.md`'s generated theme files. `docs/design/` documents
  **intent, structure, vocabulary and invariants**, and names the token or constant instead of
  reprinting a hex. This is the property the handoff lacked: the docs cannot drift from the code
[...]

Issue #348, opened 2026-08-15
CI: get cargo test --workspace actually running - missing LSP servers + a broader GPUI-window test failure


LucasPcq:
## What

`cargo test --workspace` (restored to CI in #347, previously build-only) will very likely hang on
GitHub Actions' Linux job, the same way it hung locally in this session's sandbox. Root-caused, not
guessed — this issue is the write-up plus a scoped set of next steps.

## Root cause

Five tests spawn a **real, live** `typescript-language-server`, `vue-language-server`, or
`pyright-langserver` process (verified via code inspection - not `spawn_fake_server`, which several
neighboring tests use and are unaffected):

- `crates/app/src/lsp/client.rs:4704` `restarting_a_companion_forgets_its_own_documents_without_touching_its_primary`
- `crates/app/src/lsp/client.rs:5308` `a_real_typescript_diagnostic_reaches_file_view_diagnostics_through_the_real_app_code_path`
- `crates/app/src/lsp/client.rs:5711` `typescript_language_server_tracks_a_real_live_unsaved_edit_for_both_diagnostics_and_completions`
- `crates/app/src/lsp/client.rs:5834` `pyright_tracks_a_real_live_unsaved_edit_for_both_diagnostics_and_completions`
- `crates/app/src/code_surface/lsp_ui.rs:3001` `a_real_vue_file_gets_real_diagnostics_from_both_servers_and_a_real_hover`

None of `typescript-language-server`, `@vue/language-server`, or `pyright` are installed by
`ci.yml`'s Linux job (only system dev packages for compiling GPUI's backend are installed - no
`npm i -g`/`pip install` step for any language server, not even `rust-analyzer`). When the binary
doesn't exist, the spawn attempt doesn't fail fast - it blocks. Under `cargo test`'s default
parallelism, enough of these tests each occupy a worker thread waiting on a process that will never
respond, and the *entire* suite appears frozen (observed locally: 2,808 of ~2,900 tests logged, then
0 progress for 13+ minutes before being killed).

## What was verified locally (this session)

Installed `typescript-language-server`, `typescript`, `@vue/language-server@3.3.8` (the exact
version pinned in `crate::language`'s own module docs), `pyright`, and `gopls`, then re-ran the
affected tests in isolation:

| Test | Result after installing |
|---|---|
| `a_real_vue_file_gets_real_diagnostics_from_both_servers_and_a_real_hover` | **Passes** (2.15s) |
[...]

PR #419, opened 2026-08-16
docs: sweep every design_handoff_jerry_ade citation out of the source (2/3)

LucasPcq:
Part 2 of 3 for #414. Stacked on #418 — **base is `docs/414-design-write`, not `master`**, so review the diff against that branch. Merge after #418.

## What

Removes all 364 `design_handoff_jerry_ade`/`Jerry.dc.html` citations from `crates/` (81 files), plus `Cargo.toml` and `assets/icons/README.md`. Lowers the `design_handoff_citations` ratchet from 364 to **0**.

Four rules, applied per site rather than mechanically:

| Citation carries | Action |
|---|---|
| A rule or a measured value | Keep the sentence, drop the path |
| A durable design rule | Point at `docs/design/<file>.md` instead |
| Pure archaeology — revision IDs, "verified against the mockup", the mockup's own fixture/variable names | Delete |
| A "not the mockup's fake data" disclaimer | Rephrase positively |

Concretely, the third rule is most of the volume. `(Jerry.dc.html: style-hover="background:#17191b")` next to a `token(...)` call that already contains that hex is provenance, not information. The first rule is the one worth reading the diff for — e.g. `theme::surface::SIDEBAR_STRIP` keeps *"a tab only reads as connected if the strip behind it is darker than the panel"* and loses the section number, the two intermediate hex values and the revision history around it.

Two structural notes:

- **`assets/themes/*.toml` are generated, not hand-written.** They embed `theme.rs`'s module docs verbatim, so they were regenerated through their own real path (`JERRY_REGENERATE_THEMES=1 cargo test -p app --lib builtin_themes`), not edited. Their own round-trip test confirms it.
- **`docs/design/` deliberately isn't cited everywhere a path was removed.** Most sites carried a *value*, and the new docs don't hold values — a pointer there would be a dead end. The pointers land only where a module genuinely maps to a page (`rail::state` → `rail.md`, `settings::state` → `settings.md`, `icons` → `decisions.md` §8, …).

## Why

Roughly a third of the 364 citations pointed at directories that were **never committed** — `revision 2/`, `revision 3/`, `revision 5/`, `STAGE-A-CHANGELOG.md`, `REVISION-2026-08-13.md`, `AUDIT-2026-08-13-competitive-v2.md`. An outside contributor reading `theme.rs` or `rail/render.rs` hit an unresolvable path roughly every other screen.

They are also the single largest class of comment-rule violation in the codebase: `CLAUDE.md` excludes design history, revision IDs and issue archaeology from source comments precisely because that material goes stale in place, which is exactly what happened.

## Testing

- [x] `/check` (conventions + fmt + clippy `-D warnings`) passes locally
- [x] `.claude/hooks/check-conventions.sh` passes at the new baseline (`handoff citations: 0/0`)
- [x] **Behaviour is unchanged**, verified mechanically rather than asserted:

  ```sh
  git diff -- crates/ | grep '^[-+]' | grep -v '^[-+][-+]' \
    | grep -vE '^[-+][[:space:]]*(///|//!|//)'
  ```

  Output is five test assertion messages and the one test rename (`every_mapped_slot_from_the_design_handoff_has_its_phosphor_glyph` → `every_mapped_icon_slot_has_its_phosphor_glyph`). Nothing else.
- [x] `cargo test -p wt-core` — 302 passed
- [x] `cargo test -p app --lib` on the pure modules touched heaviest: `theme::` (142), `icons::` (19), `rail::state` (43), `rail::status` (25), `sidebar::sections` (16), `sidebar::changes` (41), `palette::state` (27), `provenance::change_set` (9), `run_history::model` (27) — all pass
- [x] `cargo test -p app --lib builtin_themes` — 9 passed, confirming the regenerated theme files match their generator
- [x] `rail::strip` (5), `search::` (1) and `settings::state` (1) each have failures — **pre-existing**, reproduced identically on the parent branch before this diff, consistent with #348
- [ ] `verify` capture — n/a, comments only
- [ ] New behaviour test — n/a, no behaviour change

## Architecture notes

None.

### DIFF
--- a/.claude/conventions-baseline.json
+++ b/.claude/conventions-baseline.json
@@ -2,5 +2,5 @@
   "_comment": "Ratchet baseline for .claude/hooks/check-conventions.sh - these numbers may only go down, never up. Recount with the exact commands in that script's comments after fixing a batch of violations, and lower the number in the same commit.",
   "glob_imports_app_src": 304,
   "render_adapter_calls": 222,
-  "design_handoff_citations": 364
+  "design_handoff_citations": 0
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,8 +38,7 @@ dbg_macro = "warn"
 gpui = { git = "https://github.com/zed-industries/zed", rev = "7b030b500810b04cf5fb4aa5973be99a502d9f36" }
 anyhow = "1"
 thiserror = "1"
-# Real config-file (de)serialization for `crate::settings_store::Settings`
-# (`design_handoff_jerry_ade/revision/CHANGELOG.md`'s 2026-07-29 entry, change 3). Pinned to
+# Real config-file (de)serialization for `crate::settings_store::Settings`. Pinned to
 # the exact major versions `vendor/zed/Cargo.toml` itself pins (`serde = "1.0.221"`,
 # `toml = "0.8"`), matching this project's established "same versions Zed verified against"
 # convention for shared dependencies (see `crates/app/Cargo.toml`'s own `tree-sitter`/
--- a/assets/icons/README.md
+++ b/assets/icons/README.md
@@ -15,8 +15,8 @@ the upstream project's own `assets/bold/` tree.
 
 ## Weight
 
-All twelve files are the **bold** weight. `design_handoff_jerry_ade/revision 5/`'s
-`REVISION-2026-08-14.md` §8 states the rule verbatim:
+All twelve files are the **bold** weight. The design states the rule verbatim
+(`docs/design/decisions.md` §8):
 
 > **Weight:** `bold` at 15–17px (`regular`'s 1.5px stroke reads thin against `#5e646a`);
 > `regular` only at 20px+.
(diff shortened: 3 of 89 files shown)


LucasPcq:
One thing worth knowing while reviewing this diff, found while doing it — filed as #421 rather than folded in here.

This PR removes every **path** citation, which means every surviving `§` marker in `crates/` is now bare: `§4t`, `§2.3`, `§4u′`, `§7 rule 5`. There are **1,102 lines** carrying one, pointing at revision documents that were never in the repo.

They're overwhelmingly pre-existing, not created here:

| | lines with `§` | naming a document on the same line | bare |
|---|---|---|---|
| before this PR | 1,244 | 366 | **878** |
| after | 1,102 | 0 | 1,102 |

So 878 were already bare, and this PR *reduced* the total by 142 — but it does finish the job of stripping the last resolvable half of the remainder, which is why it's worth saying out loud rather than leaving to be discovered.

Left out deliberately: 1,102 more per-site edits would bury the property that makes this diff reviewable at all — that every non-comment line in it is a test assertion message. #421 has the scope and the same four rules.

Issue #426, opened 2026-08-16
Restore cargo test to CI on every PR and to the local pre-commit gate


LucasPcq:
## What

The last step of the epic. Once the suite is green and fast, put it back where it belongs: on every
PR in CI, and in the local pre-commit gate. Then delete every "tests are deliberately not part of
the gate — see #348" caveat scattered across the repo.

**Blocked on** the macOS process-kill fix, the nextest adoption, the testing policy, and the purge.
Do not start this until `cargo nextest run --workspace` is actually green on Linux and macOS.

### New `Linux (test)` job

`cargo nextest run --workspace --profile ci`. Tiers `unit` + `ui` only — `external` tests are
`#[ignore]`'d and skipped by default. Same system-dependency install step as the existing Linux job.
Keep the `git config --global` identity step: it is the real fix for `wt-core`'s submodule-commit
test failing with "Author identity unknown" on a fresh runner.

### New `external` job

`workflow_dispatch` + a nightly `schedule` **only — never on PRs.** Installs the pinned language
servers, then `cargo nextest run --workspace --profile ci --run-ignored all`.

Version pinning, per this project's "same version we verified against, not latest" convention
(see `crates/app/Cargo.toml`'s `tree-sitter` / `alacritty_terminal` comments for the precedent):
- `@vue/language-server@3.3.8` — already pinned in `crate::language`'s own module docs.
- `typescript-language-server`, `pyright`, `gopls` — versions still need establishing. Check
  `git blame` on each test for a version named in its commit message. #348 recorded that
  `typescript-language-server@5.3.0` ("latest") does **not** match what those tests expect, so
  "latest" is not an option.

### macOS and Windows jobs

Drop the narrow `cargo test -p app --lib status_bar::process_stats` filter for the full
`cargo nextest run --workspace` once green. The macOS job running the full suite is exactly what
would have caught the `/proc` process-tree bug.

## Restore the local gate

`cargo nextest run --workspace` joins conventions + fmt + clippy in:
- `.claude/hooks/pre-commit-check.sh` (its own header comment names #348)
- `.claude/commands/check.md`

## Remove every #348 caveat

Full list from `grep -rn "348"` on `master` today:

- [ ] `CLAUDE.md:28` — the whole "deliberately not part of this gate" paragraph
- [ ] `CONTRIBUTING.md:29`
- [ ] `README.md:116`
- [ ] `docs/development-workflow.md:52`
- [ ] `.github/pull_request_template.md:20`
[...]

Issue #440, opened 2026-08-16
Windows: the full test suite does not pass (134 failed + 26 timed out) — deferred out of #427


LucasPcq:
Portability audit for #427, ahead of #426 widening the Windows and Linux CI jobs to run the full
suite for the first time. Audit only — no code was changed, and there is deliberately no PR and no
report file: this issue is the whole deliverable.

## The question

#427's wave 2 traced **113 of 115** macOS test failures to one path-identity bug. `AdeApp`
canonicalizes the repo root it is given (`rail::repo::canonical_repo_path`,
`crates/app/src/rail/repo.rs:166`) and then keys everything by exact `PathBuf` equality or
`strip_prefix`. A fixture handing it a bare `TempDir::path()` while building its own expected paths
from that same uncanonicalized root never matches, because `/var` is a symlink to `/private/var` on
macOS.

Does that have a Windows twin, and will #426's first Windows run hit a wall?

## Short answer

**The mechanism exists on Windows and is unconditional there — but it fires on exactly the same set
of call sites wave 2 is already fixing, so it is not an additional wall.** The genuinely
Windows-only work is about 10 lines. There is also one real, user-facing Windows *product* bug that
has nothing to do with tests.

## What actually breaks path equality (verified in std 1.95.0, not assumed)

| Question | Answer | Evidence |
|---|---|---|
| How does `Path == Path` compare? | Component-wise, not byte-wise | `std/src/path.rs:3650` |
| Drive-letter case (`C:` vs `c:`)? | **Not a bug** — normalized to uppercase | `std/src/sys/path/windows_prefix.rs:137`, `parse_drive` returns `drive.to_ascii_uppercase()` |
| `/` vs `\`? | **Not a bug** — both are separators for non-verbatim paths | `windows_prefix.rs:156` |
| The `\\?\` verbatim prefix? | **This is the bug** | `path.rs:462` `PrefixComponent::eq` compares `Prefix`, which derives `PartialEq` (`path.rs:149`) — so `VerbatimDisk(b'C') != Disk(b'C')` |
| Does `Hash` agree with `Eq`? | Yes | `path.rs:3691` |

So two of the three Windows path traps usually worried about are **not** problems here. Git on
Windows emitting `C:/Users/...` compares equal to `C:\Users\...` — that is why
`wt_core::parse_worktree_list_porcelain` (`crates/wt-core/src/lib.rs:288`) getting forward slashes
from git is harmless.

[...]

Issue #427, opened 2026-08-16
Epic: test suite health — make it runnable everywhere, cut it in half, give it a policy


LucasPcq:
Umbrella issue. Tests are a brake on development in this repo, not a safety net: they take too long,
they sometimes fail and then **stay alive pegging the machine**, there is no written rule for what
deserves a test, and `cargo test --workspace` is excluded from the pre-commit gate, from `/check`,
and from CI entirely — so **nothing verifies behaviour on a PR today**.

[#348](https://github.com/ColinEspinas/jerry/issues/348) framed this as "missing language servers in
CI, plus some broad GPUI-window failure of unknown cause". Measuring it properly shows the picture is
different and worse, and the root cause of the CPU problem is a **product bug**, not test hygiene.
This issue supersedes #348, which is closed against it.

---

## What was measured

macOS 25.5, 12 cores, all language servers installed, on `master` at `35532cf`:

| | |
|---|---|
| Tests in the workspace | **3,564** (`app` 3,190 · `wt-core` 302 · `lsp-core` 53 · `pty-core` 19) |
| Lines inside `#[cfg(test)] mod` blocks | **109,987 of 242,846 — 45.3% of the workspace** |
| `#[gpui::test]` (needs a window/harness) | 1,006 |
| `thread::sleep` / wall-clock waits in tests | **303** sites, incl. `from_secs(9)` and 10× `from_millis(500)` |
| `fn git(dir, args)` helper redefined | **~30 times**, copy-pasted per test module |
| `Command::new` sites (real subprocesses) | 144 |
| tempdir sites | 1,223 |
| `app` debug test binary | 160 MB — one binary, one process, no per-test timeout |

**The core crates already fail on macOS**, with every language server installed:

- `cargo test -p pty-core` → **3 failed / 16**
- `cargo test -p lsp-core` → **14 failed / 51**
- `cargo test -p wt-core` → 302 passed in 7.9s (clean)

**The `app` suite leaks processes and threads while it runs.** Sampling the live test process
(`--test-threads=4`) at ~5 min of CPU time:

- **509 leaked child processes** and climbing — 281 `/bin/zsh`, 16 `node` (orphaned language servers)
- **630 OS threads**, of which 271 leaked `pty_core::run_writer_loop`, 196 leaked
  `pty_core::run_reader_loop`, 131 leaked `notify-rs fsevents` watcher loops
- The run **never finished** — killed manually at ~22 minutes wall-clock

## Root cause of the leak

[...]

Issue #424, opened 2026-08-16
Define a testing policy (docs/testing.md) and build the shared crates/test-support


LucasPcq:
## What

There is no written rule in this repo for what deserves a test, at what cost, or what to delete.
`CLAUDE.md`'s "Testing" section is four lines and says only *how* to write a GPUI test, never
*whether* to. That absence is why the suite grew to 3,564 tests and 109,987 lines inside
`#[cfg(test)]` blocks — **45.3% of the workspace's 242,846 lines** — with `fn git(dir, args)`
copy-pasted ~30 times and 1,223 separate tempdir setups.

This issue writes the policy and builds the shared test infrastructure the policy assumes. It
deletes nothing — it is what the purge issue is executed *against*.

### Three tiers

| Tier | Marker | May do | Must not | Budget |
|---|---|---|---|---|
| `unit` | plain `#[test]` | pure logic, tempdir filesystem | spawn a process, open a gpui window, sleep, assert on wall-clock | < 10 ms |
| `ui` | `#[gpui::test]` + `TestAppContext` / `VisualTestContext` | real git fixtures, real key/action dispatch | external language servers, real agent processes, sleep | < 2 s |
| `external` | `#[ignore = "external: <binary>; see docs/testing.md"]` | real `rust-analyzer` / `typescript-language-server` / `pyright` / `gopls`, real OS process-tree semantics | — | dedicated CI job only |

Tiers 1 and 2 are the PR gate. Tier 3 never is.

Today that is 2,186 plain `#[test]` and 1,006 `#[gpui::test]` in `crates/app` alone — the split
already roughly exists, it has just never been named or enforced.

### What deserves a test

- One test per behaviour a **user** could notice — not one per branch of the implementation.
- One regression test per fixed bug, named after the symptom.
- Pure derivations (colours, geometry, formatting) get **one table-driven invariant test**, not one
  test per input.
- Explicitly not worth a test: a constant equalling its own literal; a field round-tripping through
  its own getter; "a render function returned some element"; exhaustively enumerating a sweep the
  code itself already iterates over.

### Deletion criteria

These are the five the purge issue cites, one per removal, in its PR bodies:

1. **Sweep redundancy** — N tests differing only by an input value. Collapse to one table-driven test.
2. **Implementation mirroring** — the assertion recomputes what the code under test does.
3. **Vacuous** — would still pass with the feature removed, or only asserts on test-local data.
[...]

PR #444, opened 2026-08-16
chore(ci): restore the test suite to CI and the local pre-commit gate (#426)

LucasPcq:
## Summary

Closes #426 — the last piece of epic #427. Puts the test suite back where it belongs: on every PR
in CI (Linux and macOS), in the local pre-commit gate, and gives the `external` tier its own
nightly job. Windows builds and runs its platform-specific sampling tests only — see "Wave 3"
below for why that is deliberate. Then deletes every "tests are deliberately not part of the gate — see #348"
caveat in the repo.

**Status: wave 3 applied — see the section at the bottom, which supersedes parts of what
follows.** The original body is kept intact below for review history; where it conflicts with
"Wave 3", wave 3 is current. The suite is green locally and the first real CI run has happened,
so the "nothing has ever run" caveats below are no longer true.

## Why

The suite left CI in #348 because one test blocking on a language server that never answers froze
the whole run with no way to attribute it (2,808 of ~2,900 tests logged, then 13+ minutes of
nothing). Every part of that has since been fixed by this epic: nextest gives per-test processes
and a real timeout, and `docs/testing.md`'s three tiers say which tests may spawn a real server at
all. Nothing is left justifying the caveats.

### `.github/workflows/ci.yml`

- **New `Linux (test)` job** — `cargo nextest run --workspace --profile ci`, the `unit` + `ui`
  tiers (`external` is `#[ignore]`d and skipped by default). Same system-dependency install as the
  existing Linux job.
- **Restored the `git config --global` identity step.** The comment in `ci.yml` claimed this step
  existed; it did not — it was deleted along with the test step in d30e7de. It is the real fix for
  `wt-core`'s
  `discard_worktree_refuses_when_stash_push_captures_nothing_and_an_unrelated_stash_already_exists`
  failing with "Author identity unknown" on a fresh runner (`git submodule update --init` creates
  an independent git dir that does not inherit the fixture's *local* identity). It is now on all
  three test jobs, not just Linux.
- **New `External (real language servers)` job** — `workflow_dispatch` + a nightly `schedule`
  only, never on a PR, enforced by an explicit `if:` on the job. Installs the servers, then
  `--profile external --run-ignored all`.
- **macOS widened** from `cargo test -p app --lib status_bar::process_stats` to the full
  `cargo nextest run --workspace --profile ci`. (Windows was widened too, then reverted in wave 3 —
  see below.)
- `cargo-nextest` is installed via `taiki-e/install-action@v2`, pinned to `0.9.143` — the version
  this change was written and validated against, per the project's "same version we verified
  against, not latest" convention.

### Language-server version pinning

This was the real open work #426 left. What I could establish, and what I could not:

| Server | Pin | Evidence |
|---|---|---|
| `rust-analyzer` | toolchain `1.95.0` | It is a **rustup component**, not a separate download. `components: rust-analyzer` on `dtolnay/rust-toolchain` pins it to the same toolchain as everything else — the strongest pin available, and free. Verified: `rustup component list` reports `rust-analyzer-<host>`, and installing it puts a `rust-analyzer` proxy on `PATH`. |
| `@vue/language-server` | **`3.3.8`** | `crate::language`'s module docs record the whole live investigation against exactly that version, and #348 re-confirmed the Vue test passes against it (2.15s). |
| `typescript-language-server` | **unpinned, knowingly** | `git blame` on all three of its tests leads to b0c08523 / e75f9dd3 / 06b06dac / 1aa4cd54; **not one names a version**, in the commit message or in a nearby comment, and nothing else in the repo records one. The only established fact is negative: #348 found `5.3.0` ("latest") fails them. |
| `pyright` | **unpinned, knowingly** | Same test, same blame (b0c08523), same result: no version recorded anywhere. #348 saw "latest" fail with the same shape. |
| `gopls` | **not installed** | Not a gap — a correction. `crate::language`'s registry gives Go `lsp: None` and `gopls` is PATH-detection-only on the Settings page ("a real detectable binary this app deliberately never spawns a client for"). **No test in this workspace spawns it.** #426 inherited it from #348's list; installing it would be cargo-culting. |

Rather than guess a pin — a wrong pin looks authoritative and hides the drift — the external job
has a `Resolved server versions` step that prints `npm ls -g --depth=0` and `rust-analyzer
--version` on every nightly run. That turns the nightly into the mechanism that *produces* the
evidence for a real pin, instead of a place a guess would quietly rot. `TODO(#426)` marks it.

### `.config/nextest.toml` — the TODO(#424) the tier issue left

Resolved as a `[profile.external]` (`inherits = "ci"`, `slow-timeout` 120s) rather than the
`[[profile.default.overrides]]` + filterset the TODO imagined, **because that filterset cannot be
written**: nextest's filterset language has no ignore-status predicate. Verified empirically
against cargo-nextest 0.9.143 — `ignored()` is a parse error, and `default()` refers to the
profile's `default-filter`, not to `#[ignore]`. The alternative, enumerating the tier's test names
in a `test(/.../)` filterset, would rot silently the first time a tier member is added or renamed.
Selecting the tier by the profile its own CI job runs under is the version of this that cannot
drift. The reasoning is in the file, not just here.

### The local gate

`cargo nextest run --workspace` joins conventions + fmt + clippy in
`.claude/hooks/pre-commit-check.sh` and `.claude/commands/check.md`. The hook's exit contract is
preserved: a missing `cargo-nextest` skips the step (exit 0, allow) rather than blocking a
contributor who has not run `/setup` — only a genuine test failure denies.

### Every #348 caveat removed

Re-derived with a fresh `grep -rn "348" .` rather than trusting #426's list, which is stale (it
names `.claude/skills/setup/SKILL.md`, which does not exist — the setup entry point is
`.claude/commands/setup.md` — and predates `docs/testing.md`, which carries its own #348
paragraph). `Cargo.lock` and `design_handoff_jerry_ade/*.html` match incidentally.

Removed: `CLAUDE.md`, `CONTRIBUTING.md`, `README.md`, `docs/development-workflow.md`,
`docs/testing.md`, `.github/pull_request_template.md`, `.claude/skills/ship/SKILL.md`,
`.claude/hooks/pre-commit-check.sh`, `.claude/commands/check.md`, `.github/workflows/ci.yml`'s
job comment.

Four `#348` references remain **on purpose**, and none is a caveat — they are the citation for a
factual claim whose only record is that issue: the observed hang that justifies nextest over
`cargo test`, and the "5.3.0 fails" finding that is the entire reason the two npm servers are
unpinned. Deleting them to satisfy a grep would destroy the evidence the next person needs. Flagging
this explicitly since #426's "Done when" asks for a clean grep — say the word and I will inline the
facts and drop the numbers.

## Testing

- [x] `.claude/hooks/check-conventions.sh` — OK (globs 301/301, render adapter calls 201/201 after
      the ratchet commit; see the caveat about `CLAUDE_PROJECT_DIR` below)
- [x] `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings` — no
      Rust changed in this PR
- [x] `actionlint` on `.github/workflows/ci.yml` — exit 0, clean (includes its shellcheck pass over
      every `run:` block)
- [x] YAML parsed and every job/step/`if:`/trigger asserted structurally
- [x] `.config/nextest.toml` — `[profile.external]` proven to parse, inherit `[profile.ci.junit]`
      (a real `junit.xml` is written to `target/nextest/external/`), and work in this workspace via
      `cargo nextest run -p test-support --profile external --run-ignored all`
- [x] `cargo nextest run -p wt-core -p test-support -p pty-core -p lsp-core --profile ci` — **392
      passed, 0 failed**
- [x] `.claude/hooks/pre-commit-check.sh` — `bash -n` clean; non-`git commit` input still exits 0
- [x] `cargo nextest run --workspace --profile ci` — ran to completion on macOS:
      **3,341 tests in 167s (2m47s); 3,216 passed, 125 failed, 9 skipped.** Red, as expected — the
      125 are the D5/D6/D7 slices still in flight, not this PR (which changes no Rust). Two things
      this run does establish:
      - **The runtime target is realistic.** 2m47s locally, against a ~10-minute CI budget.
      - **The `external` tier is genuinely skipped by default** — 9 skipped, 0 attempted spawns, no
        hang. That is the exact failure mode #348 was about, and it is gone.

### One real find already, from running on macOS

`app keymap::tests::resolve_keystroke_renders_the_secondary_modifier_as_the_cross_platform_mod_glyph`
fails at `crates/app/src/keymap.rs:328` with `left: ["⌃", "N"]` / `right: ["⌘", "N"]` — the
secondary-modifier glyph resolves to Control where the test expects Command, **on a real macOS
machine**. This is precisely the class of bug the narrow `status_bar::process_stats` filter could
never have caught, found the first time the suite was pointed at a Mac. It is in `crates/app` and so
outside this PR's boundary; flagging it for whoever owns that slice rather than fixing it here.

### What can only be verified once the suite is green

Stated plainly rather than claimed:

- **That the jobs pass.** Every job's *syntax and structure* is verified; none of them has run.
- **macOS and Windows.** Nobody has ever run this suite on Windows, and macOS has only ever run one
  narrow module filter. The bug that dominated this epic was macOS-specific path canonicalization
  (`/var` → `/private/var`); Windows has its own untested path-equality traps (drive-letter case,
  UNC, `\\?\`). **Expect the first real run to find things — that is this change working, not this
  change failing.**
- **The ~10-minute Linux target.** Unmeasurable until it runs on a real runner.
- **The external job end to end.** It needs a live npm registry and real server processes.
- **Whether the two unpinned servers pass at all.** That is exactly what the nightly is for.

## The ratchet (separate commit, 731e33c)

`render_adapter_calls` lowered 221 → 201, the real count on this branch. The epic's purge slices
already removed those 20 and lowered `glob_imports_app_src` in the same pass, but left this number
where it was — and a ratchet sitting 20 above the truth silently permits 20 new violations, which
is the one thing it exists to prevent. Kept as its own commit because this file is a conflict
magnet while D5/D6/D7 land in parallel; drop or redo it on rebase without touching the rest.

**A real trap found while doing it, worth knowing about:** `check-conventions.sh` starts with
`cd "${CLAUDE_PROJECT_DIR:-.}"`. Run from a worktree while `CLAUDE_PROJECT_DIR` still points at the
main checkout, it silently measures **the wrong tree** and reports a confident, wrong pass. It told
me `201/221` and then `222/222` for the same command in the same directory, because it was reading
`~/Documents/Dev/jerry` (on `master`, 304/222) both times, not this branch. Verified the real
numbers by running the greps directly here (301 globs, 201 adapter calls) and re-running the script
with `CLAUDE_PROJECT_DIR` pointed at the worktree: `OK (glob imports: 301/301, render adapter
calls: 201/201)`. Anyone rebasing this file should do the same, or they will confirm a number that
was never measured.

## Notes for the reviewer

- **The gate blocked its own commit**, exactly as designed: the extended
  `.claude/hooks/pre-commit-check.sh` denied `git commit` because the suite is red from the
  in-flight slices. Both commits were made through a one-time, explicitly authorized bypass of that
  hook, with the gate itself left exactly as written. Recording it here rather than quietly routing
  around it.
- The purge is still in flight, so `crates/` is untouched by this PR by design — including
  `crates/app/src/lsp/client.rs`'s four remaining bare `#[ignore]`s (D6's slice). Their `#348`
  comments and missing reason strings are that PR's job, not this one's.
- Observed while running the core crates: `lsp-core`'s
  `a_real_but_frozen_server_fails_the_write_within_the_budget_instead_of_hanging_forever` and
  `a_writer_queued_behind_one_that_gives_up_is_refused_rather_than_corrupting_the_stream` each take
  **30.045s** and trip nextest's SLOW threshold. They pass, and the duration is the budget they are
  asserting — but they are `unit`-tier tests 3,000× over that tier's 10 ms budget, and together
  they are a minute of every CI run. Worth a look as a follow-up; deliberately not touched here.

---

# Wave 3 — the first real CI run, and what it found

The first run of the jobs above is what this section is about. It supersedes the "what can only be
verified once the suite is green" caveats: the suite has now run.

## The four failures, and why they are one defect

Linux reported 4 failures out of 3128; macOS 4 out of 3125. They are the same four, and all four
are an `external`-tier test wearing a `ui`-tier marker — the exact case `docs/testing.md`'s tier
table exists to separate.

| Test | Needs | Now |
|---|---|---|
| `code_surface::file_view::lsp_failed_status_chip_tests::clicking_the_failed_status_chip_really_restarts_the_language_servers` | a real `rust-analyzer` | `#[ignore = "external: rust-analyzer; …"]` |
| `rail::render::repo_checkout_tests::build_repo_groups_folds_a_non_focused_repos_real_agent_into_its_own_row` | the `claude` CLI | `#[ignore = "external: claude; …"]` |
| `root::repo_list_tests::checkout_repo_from_rail_leaves_the_previous_repos_agents_running` | the `claude` CLI | `#[ignore = "external: claude; …"]` |
| `sidebar::render::changes_sections_tests::a_live_runs_row_states_that_it_is_running_and_switching_focus_changes_no_count` | the `claude` CLI | `#[ignore = "external: claude; …"]` |

The reason strings name the specific binary rather than a generic "external": an agent CLI is not
a language server, and a future reader installing the wrong one learns nothing.

**Read rather than assumed.** `ProcessKind::claude()` resolves through `AgentKind::binary_name()`
to the literal `claude` on `$PATH`, with no fallback, and all three of those tests then assert the
process is genuinely alive (`Status::Run`, `TerminalPane::is_running`, `row.live`). The CI log is
explicit about the consequence: `assertion left == right failed … left: Fail, right: Run`. The PTY
child exec'd a binary that does not exist on a runner and exited immediately. The LSP one is the
same shape one layer over: it asserts the click's own `ensure_lsp_client` reached
`Spawning`/`Ready`, which cannot happen without `rust-analyzer` installed.

These pass on the maintainer's laptop only because that laptop has both binaries — verified, the
real `claude` there is `~/.superconductor/bin/claude`. "Passes where the binary happens to exist,
fails on a clean runner" is the definition of the `external` tier.

## The fifth test: no defect, and no change

`graph_view::render::graph_virtualization_tests::loading_more_commits_keeps_the_selected_row_on_the_same_commit`
was reported as a fifth, Linux-only failure. **It is not a failure.** The Linux job log records:

```
PASS [   6.852s] ( 811/3128) app graph_view::render::graph_virtualization_tests::loading_more_commits_keeps_the_selected_row_on_the_same_commit
…
Summary [ 426.154s] 3128 tests run: 3124 passed, 4 failed, 30 skipped
```

Four failures, and this is not one of them. The 6.9s is its duration, not a timeout — nextest's
slow threshold is 30s. That time is real fixture cost (`seed_commits` writing 520 real git commits
through the `git` CLI), not a timing assumption, so there is no `run_until_parked`/`wait_until` fix
to make and nothing to re-tier. Deliberately left alone.

## Closing the class, not the four instances

Eleven other tests spawn a real agent CLI and none of them failed, which is the interesting case:
a grep cannot tell a test that *needs* the binary from one that merely *calls* the API.

So this was verified by experiment rather than by reading. The entire workspace was run with
`claude`, `codex`, `rust-analyzer`, `typescript-language-server`, `pyright` and
`vue-language-server` all removed from `PATH` (a shim `PATH` keeping `cargo`/`rustc`/`node`, since
the CI runner does have `node` — several `ui` tests legitimately drive a `node`-based *fake* LSP
server, which is not an external server and correctly stays in the gate):

```
Summary [ 113.861s] 3121 tests run: 3121 passed, 34 skipped
```

Zero failures with every external binary absent. The eleven remaining agent-spawning tests assert
on tab-strip and rail *structure*, not on process liveness, so they behave identically either way —
hermetic, and they stay in the gate. The class is closed with evidence, not with a blanket re-tier
that would have cost eleven green tests for nothing.

## Windows: back to its narrow scope (issue #440)

The one full-suite Windows run scored **134 failures and 26 timeouts across 3073 tests, in 1407
seconds**. That is a real investigation — paths formatted into strings, forward slashes baked into
expected values, and 26 timeouts whose mechanism nobody has diagnosed — and it is #440's work, split
out of #427 rather than allowed to hold a finished epic behind unbounded work.

So the Windows job is reverted to exactly what it ran before this PR: `cargo build --workspace`
plus `cargo test -p app --lib status_bar::process_stats`, the `GetProcessTimes`/PSAPI FFI backend
that is covered nowhere else. The job carries a `DELIBERATELY NARROW — do not widen this without
reading issue #440 first` comment stating *why*, so the next reader does not take it for an
oversight. `docs/testing.md` and `README.md` no longer claim the suite runs on Windows; testing.md
now says plainly that Windows is not a test platform yet and that a Windows result is not a gate.

Windows coverage is deferred, not forgotten.

## Duplicate runs: about 200 billed minutes back per push

The workflow triggered on bare `push:` **and** `pull_request:`, so every push to a PR branch fired
two complete, identical runs. Verified on this branch — runs `31974623400` (push) and
`31974624833` (pull_request) are the same commit, and there were three such pairs. At GitHub's
billing multipliers (Linux 1×, Windows 2×, macOS 10×) one run of this workflow is roughly 200
billed minutes, so the duplicate cost about that much per push for no added signal.

- `push:` is now scoped to `branches: [master]`. PR branches run once, via `pull_request`.
  `workflow_dispatch` and `schedule` are untouched — the `external` job depends on them.
- A `concurrency:` group with `cancel-in-progress: true` now supersedes a run still in flight
  instead of racing it.
- The comment above the triggers claimed both events existed for the `external` job. That was not
  accurate — `external` is gated on `workflow_dispatch` and `schedule` — and it is corrected rather
  than preserved.

**One deliberate refinement over the literal ask:** the concurrency key includes `github.event_name`
alongside workflow and ref. Without it, a push to `master` and the nightly `schedule` run share a
ref and would cancel each other — and the nightly is the only execution the `external` tier ever
gets. Flagging it explicitly since it is a small deviation.

**Verified on the very first push:** commit `bb3396c` produced exactly one run, where every
preceding commit on this branch produced a push/pull_request pair.

## macOS cost — noted, not acted on

macOS is the most expensive job here at 10× billing, and its scope is deliberately unchanged (it is
the platform this epic's real product bug lived on). Two cheap, scope-preserving ideas, raised here
rather than done:

1. **The `cargo build --workspace` step is likely redundant with the test step.**
   `cargo nextest run --workspace` compiles the same dependency graph plus the test harnesses, into
   the same target dir. Dropping the explicit build would not reduce what is *tested*, only what is
   compiled twice — but it would stop catching a break in a non-test target, so it wants measuring
   before anyone does it.
2. **`Swatinem/rust-cache` currently saves the cache on every run.** Adding
   `save-if: ${{ github.ref == 'refs/heads/master' }}` would keep the restore on PR runs while
   skipping the compress-and-upload step, which is pure overhead on a 10×-billed runner.

## Verification

- [x] `CLAUDE_PROJECT_DIR=$PWD .claude/hooks/check-conventions.sh` — `OK (glob imports: 300/300,
      render adapter calls: 196/196)`
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo nextest run --workspace` — **3121 passed, 34 skipped, 0 failed.** Was 3125 passed / 30
      skipped; exactly the four re-tiered tests moved, and nothing else changed.
- [x] The same run with every external binary stripped from `PATH` — **3121 passed, 34 skipped, 0
      failed** (see "Closing the class").

### DIFF
--- a/.claude/commands/check.md
+++ b/.claude/commands/check.md
@@ -1,5 +1,5 @@
 ---
-description: Run the full pre-commit gate (conventions, fmt, clippy) - the same thing CI runs
+description: Run the full pre-commit gate (conventions, fmt, clippy, tests) - the same thing CI runs
 model: haiku
 ---
 
@@ -10,14 +10,18 @@ rather than continuing past it:
 .claude/hooks/check-conventions.sh
 cargo fmt --all -- --check
 cargo clippy --workspace --all-targets -- -D warnings
+cargo nextest run --workspace
 ```
 
 If a step fails, show the actual failing output (not a summary) and stop — don't run later steps
 against code that hasn't passed the earlier ones. If everything passes, say so in one line; this
 command doesn't need a report beyond pass/fail plus whatever failed.
 
-**`cargo test --workspace` is deliberately not part of this gate right now** — see
-[GitHub issue #348](https://github.com/ColinEspinas/jerry/issues/348), which tracks getting the
-suite back into CI and this gate once resolved. If your change needs test coverage verified, run
-the relevant scoped tests yourself (e.g. `cargo test -p wt-core`, or `cargo test -p app --lib
-<module>::`) — don't treat a clean `/check` as proof the full suite passes.
+The test step covers the `unit` and `ui` tiers, which is what CI's `Linux (test)` job runs. The
+`external` tier is `#[ignore]`d and skipped — it needs real language servers and runs only in the
+nightly `External` job (see [`docs/testing.md`](../../docs/testing.md)). If you changed something
+that tier covers, run it yourself with `cargo nextest run --workspace --profile external
+--run-ignored all` and the servers installed.
+
+If `cargo nextest` isn't installed, run `/setup` — don't fall back to `cargo test`, which has no
+per-test timeout and will simply hang on a blocked test instead of naming it.
--- a/.claude/conventions-baseline.json
+++ b/.claude/conventions-baseline.json
@@ -1,5 +1,5 @@
 {
   "_comment": "Ratchet baseline for .claude/hooks/check-conventions.sh - these numbers may only go down, never up. Recount with the exact commands in that script's comments after fixing a batch of violations, and lower the number in the same commit.",
   "glob_imports_app_src": 301,
-  "render_adapter_calls": 221
+  "render_adapter_calls": 201
 }
--- a/.claude/conventions-baseline.json
+++ b/.claude/conventions-baseline.json
@@ -1,5 +1,5 @@
 {
   "_comment": "Ratchet baseline for .claude/hooks/check-conventions.sh - these numbers may only go down, never up. Recount with the exact commands in that script's comments after fixing a batch of violations, and lower the number in the same commit.",
-  "glob_imports_app_src": 301,
-  "render_adapter_calls": 201
+  "glob_imports_app_src": 300,
+  "render_adapter_calls": 196
 }
--- a/.claude/hooks/pre-commit-check.sh
+++ b/.claude/hooks/pre-commit-check.sh
@@ -1,14 +1,17 @@
 #!/usr/bin/env bash
 # PreToolUse:Bash hook. Runs CLAUDE.md's pre-commit gate - the structural ratchet
-# (check-conventions.sh), fmt --check, and clippy -D warnings - before a `git commit` command
-# executes, so a lint-failing or convention-regressing change never lands in history.
-# `cargo test --workspace` is intentionally not part of this gate (or `/check`'s) right now -
-# GitHub issue #348 tracks why and getting it back. Run tests relevant to your change manually.
+# (check-conventions.sh), fmt --check, clippy -D warnings, and the `unit`/`ui` test tiers - before
+# a `git commit` command executes, so a lint-failing, convention-regressing or test-breaking
+# change never lands in history.
+#
+# Tests run through `cargo nextest`, not `cargo test`: per-test processes plus
+# `.config/nextest.toml`'s `slow-timeout` mean one hung test is a named failure rather than a hook
+# that never returns. The `external` tier is `#[ignore]`d and skipped by default (docs/testing.md).
 #
 # Exit code contract, matching Claude Code's PreToolUse protocol: this hook must NEVER hard-fail
 # in a way that blocks work for a reason unrelated to the gate itself. Missing `jq`, missing
-# `cargo`, or a command that isn't a git commit all exit 0 (allow, unmodified) rather than error.
-# Only a genuine fmt/clippy failure produces a deny decision.
+# `cargo`, missing `cargo-nextest`, or a command that isn't a git commit all exit 0 (allow,
+# unmodified) rather than error. Only a genuine fmt/clippy/test failure produces a deny decision.
 
 set -u
 
@@ -36,10 +39,17 @@ fi
 if [ -z "$FAIL" ] && command -v cargo &>/dev/null; then
   cargo clippy --workspace --all-targets -- -D warnings >/tmp/jerry-precommit-clippy.log 2>&1 || FAIL="clippy"
 fi
+# `cargo-nextest` is a separate binary, not part of a rustup toolchain: a contributor who hasn't
+# run `/setup` yet skips this step rather than being blocked by it, per this file's exit contract.
+if [ -z "$FAIL" ] && command -v cargo-nextest &>/dev/null; then
+  cargo nextest run --workspace >/tmp/jerry-precommit-nextest.log 2>&1 || FAIL="nextest"
+fi
 
 if [ -n "$FAIL" ]; then
   if [ "$FAIL" = "conventions" ]; then
     REASON="the convention ratchet failed - see /tmp/jerry-precommit-conventions.log."
+  elif [ "$FAIL" = "nextest" ]; then
+    REASON="cargo nextest run --workspace failed - see /tmp/jerry-precommit-nextest.log. Run /check before committing."
   else
     REASON="cargo $FAIL failed - see /tmp/jerry-precommit-$FAIL.log. Run /check before committing."
   fi
--- a/.claude/hooks/pre-commit-check.sh
+++ b/.claude/hooks/pre-commit-check.sh
@@ -41,8 +41,15 @@ if [ -z "$FAIL" ] && command -v cargo &>/dev/null; then
 fi
 # `cargo-nextest` is a separate binary, not part of a rustup toolchain: a contributor who hasn't
 # run `/setup` yet skips this step rather than being blocked by it, per this file's exit contract.
-if [ -z "$FAIL" ] && command -v cargo-nextest &>/dev/null; then
-  cargo nextest run --workspace >/tmp/jerry-precommit-nextest.log 2>&1 || FAIL="nextest"
+# The skip is announced, not silent - a gate that quietly drops its only behavioural check reads as
+# a passing gate.
+if [ -z "$FAIL" ] && command -v cargo &>/dev/null; then
+  if command -v cargo-nextest &>/dev/null; then
+    cargo nextest run --workspace >/tmp/jerry-precommit-nextest.log 2>&1 || FAIL="nextest"
+  else
+    echo "pre-commit-check: cargo-nextest missing - tests NOT run for this commit." >&2
+    echo "  cargo install cargo-nextest --locked   (or run /setup)" >&2
+  fi
 fi
 
 if [ -n "$FAIL" ]; then
--- a/.claude/skills/ship/SKILL.md
+++ b/.claude/skills/ship/SKILL.md
@@ -10,12 +10,12 @@ changed and why from the diff alone.
 
 ## Steps
 
-1. **Run the gate.** `/check` (conventions + fmt + clippy `-D warnings`). Don't proceed past a
-   failure — a PR opened against a red gate just moves the failure to someone else's screen.
-   `cargo test --workspace` isn't part of this gate right now
-   ([issue #348](https://github.com/ColinEspinas/jerry/issues/348)) — run whatever tests are
-   relevant to the change manually (e.g. `cargo test -p wt-core`, or a scoped `cargo test -p app
-   --lib <module>::`) and note the result in the PR's Testing section instead of leaving it blank.
+1. **Run the gate.** `/check` (conventions + fmt + clippy `-D warnings` + `cargo nextest run
+   --workspace`). Don't proceed past a failure — a PR opened against a red gate just moves the
+   failure to someone else's screen. That covers the `unit` and `ui` tiers; if the change touches
+   the `external` tier (real language servers, `docs/testing.md`), run that yourself with
+   `cargo nextest run --workspace --profile external --run-ignored all` and note the result in the
+   PR's Testing section instead of leaving it blank.
 
 2. **Decide if this is UI-visible.** `git diff --name-only` against the target branch — did it
    touch a `render.rs`, `theme.rs`, or anything else that changes what the app looks like? If so,
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -12,12 +12,28 @@ slow-timeout = { period = "30s", terminate-after = 4 }
 # Flakes get fixed, not retried into a green run.
 retries = 0
 
-# TODO(#424): once the `external` test tier exists, add a `[[profile.default.overrides]]` giving it
-# a longer slow-timeout — a real language-server handshake legitimately takes seconds.
-
 [profile.ci]
 # Failures are repeated at the end of the log, where the Actions summary is readable.
 failure-output = "immediate-final"
 
 [profile.ci.junit]
 path = "junit.xml"
+
+# The `external` tier (docs/testing.md): real `rust-analyzer` / `typescript-language-server` /
+# `vue-language-server` / `pyright-langserver` processes. A real handshake plus a first index pass
+# legitimately takes seconds, so 30s/2min is too tight to distinguish "slow server" from "hung
+# test" here.
+#
+# This is a profile rather than the `[[profile.default.overrides]]` with a filterset that this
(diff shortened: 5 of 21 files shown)


LucasPcq:
### #348 verification, re-run after D6 landed

D6's slice merged as c09919e while this PR was being written. Re-checked rather than assumed:

`crates/app/src/lsp/client.rs` on `chore/test-suite-health` now has **six** properly-tiered
markers and **zero** `#348` references:

```
4670:    #[ignore = "external: vue-language-server; see docs/testing.md"]
5100:    #[ignore = "external: rust-analyzer; see docs/testing.md"]
5207:    #[ignore = "external: typescript-language-server; see docs/testing.md"]
5332:    #[ignore = "external: rust-analyzer; see docs/testing.md"]
5583:    #[ignore = "external: typescript-language-server, npm; see docs/testing.md"]
5707:    #[ignore = "external: pyright-langserver; see docs/testing.md"]
```

`crates/app/src/code_surface/lsp_ui.rs` (D2, merged earlier) likewise: four `external:` markers, no
`#348`.

So `git grep -n "348" origin/chore/test-suite-health -- . ':!Cargo.lock' ':!design_handoff_jerry_ade'`
returns **11 hits, and all 11 are in the 10 files this PR rewrites** — nothing left in `crates/`:

```
.claude/commands/check.md:20
.claude/hooks/pre-commit-check.sh:6
.claude/skills/ship/SKILL.md:16
.github/pull_request_template.md:20
.github/workflows/ci.yml:24, :25
CLAUDE.md:28
CONTRIBUTING.md:35
README.md:116
docs/development-workflow.md:52
docs/testing.md:132
```

After this merges, the only `348` matches in the tree are `Cargo.lock` checksums,
`design_handoff_jerry_ade/*.html` (both incidental), and the four deliberate evidence citations
described in the PR body.

### Base

Branched from 44bd769, one commit behind `chore/test-suite-health` (c09919e). Deliberately not
rebased yet — D5/D7 are still landing, and `.claude/conventions-baseline.json` (its own commit,
731e33c) is the file that will conflict. Rebase once wave 2 is complete, and re-derive the ratchet
number with `CLAUDE_PROJECT_DIR` actually pointed at the worktree.

PR #458, opened 2026-08-17
fix(app): scroll the command palette to follow the keyboard selection (#413)

### LINKED ISSUE #413: Palette does not scroll when moving down/up with arrow
## What happened

When using the palette with a keyboard the palette does not scroll to keep the selected item in view.

## Where

Command palette

## Reproduction

Open the command palette with a few terminal or agents open so we need to scroll, then move down with arrows.

## Environment

- OS: Windows 11
- Jerry version / commit: 0.1.0

ColinEspinas:
Closes #413

## What

Three things, in the order they were found.

**1. The palette didn't scroll to its keyboard selection.** Arrow keys moved the highlight without
moving the results viewport, so once the list overflowed the selection walked off the bottom edge
and `⏎` ran a row the user could not see. `move_palette_selection` now reveals the selected row
through the palette's existing `ScrollHandle`, using GPUI's minimum-movement strategy — a no-op when
the row is already visible, so it never fights the mouse wheel.

Making the row addressable is the substantive part. `ScrollHandle::scroll_to_item` indexes the
scroller's *direct* children, and `render_palette_groups` nested each group's rows inside a
per-group wrapper — so the handle only ever saw groups. Headers and rows are now one flat run of
children, and the new pure `palette::row_child_index` maps a flattened row index onto its child
position by skipping one child per header above it.

**2. That flattening compacted the list** (caught in review). As direct children of a column flex
container, headers and rows shrink by default, so an overflowing list squashed its rows to 19.5px
instead of `theme::band::PALETTE_ROW`'s 30px — the per-group wrapper had been absorbing that shrink.
Both now carry `flex_none`, matching `completion_popup`'s rows.

**3. Palette input lag and a key-repeat stall** (also from review: holding `↑`/`↓` froze the
highlight, which then jumped on release). `build_palette_groups` runs from both
`move_palette_selection` and `render_palette`, so anything O(files) there is paid twice per
keystroke; on a large repo that exceeded the frame budget and key-repeat events queued faster than
frames could be produced. Two costs, both in the GPUI-free half:

- `substring_match` allocated two folded `Vec<char>` per call, once per candidate *per aux field*.
  It now compares in place, with a byte-wise fast path for pure-ASCII input (file names, command
  labels, branches) — for ASCII a byte index is a char index, so the returned span is unchanged.
- `finish_group` built a `PaletteEntry` (cloned path, formatted secondary, label split into three
  owned strings) for every match, then discarded all but `MAX_ENTRIES_PER_GROUP`. It now ranks and
  caps first, and builds only the survivors.

Measured on `build_groups`, **`--release`**, 20k files:

| query | before | after | |
|---|---|---|---|
| broad (`"file"`) | 8.13ms | 0.60ms | **13.6×** |
| narrow (`"file123"`) | 1.02ms | 0.71ms | 1.4× |
| no match (`"zzz"`) | 1.10ms | 0.91ms | 1.2× |
| empty | 17.0µs | 17.5µs | — |

Worst case is now ~0.9ms per build, ~1.8ms per keystroke, inside a 16ms frame. Note these are
release figures on purpose — debug-build numbers pointed at the wrong bottleneck entirely, exactly
as `CLAUDE.md` warns.

Worth recording: this is **not** a missing virtualized list. Each group is capped at
`MAX_ENTRIES_PER_GROUP` (8), so at most ~40 rows ever render regardless of repo size — row count was
never the cost.

## Testing

- [x] `/check` — `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D
      warnings` clean. The conventions script itself skipped (`jq` not installed on this machine),
      so both counts were run by hand: `glob_imports_app_src` 300 and `render_adapter_calls` 170,
      exactly at baseline.
- [ ] `verify` capture — **not run**, by agreement: a release build plus a live window was not worth
      it here, and this machine's GPUI/PTY tests are unreliable (see below).
- [x] New/changed behavior has a real test, in the right tier:
      - `ui` — `arrowing_down_keeps_the_highlighted_row_inside_the_viewport` and
        `arrowing_back_up_brings_the_viewport_back_with_it` walk the whole list with real keystrokes
        and assert after *every* step that the highlighted row's painted box sits entirely inside
        the results viewport's. Both confirmed to fail without the scroll call.
      - `ui` — `an_overflowing_list_keeps_every_row_at_its_full_themed_height` asserts every row's
        painted height against the theme constant; fails at 19.5px without `flex_none`.
      - `unit` — `row_child_index` (header offset, strictly increasing against `flatten`,
        past-the-end/empty).
      - `unit` — for the perf change, behaviour is guarded rather than assumed: the
        char-index-vs-byte-index contract the folded buffers used to provide (**nothing covered this
        before — every existing case was ASCII**), the tail-overrun case an in-place comparison can
        get wrong, that a capped group builds only the rows it keeps, and that equal ranks keep
        candidate order. No timing assertions — those would be flaky.
- [ ] `external` tier — not touched.

`cargo nextest run --workspace` on this machine reports pre-existing, environmental failures
unrelated to this branch, concentrated in `pty-core` (real PTY spawning), `lsp-core` (path
canonicalization) and agent/session-restore tests — including crates this diff does not touch at
all. The four `tabless_window_keybinding_tests` timeouts were confirmed to reproduce identically on
a clean, stashed tree. No palette test fails.

## Architecture notes

`row_child_index` and both perf changes live in `crates/app/src/palette/state.rs`, the palette's
GPUI-free half, so the layout contract and the matching pipeline stay unit-testable without a live
window. No crate boundary or Command/Query shape changes.

## Still open

`build_palette_groups` is still called twice per keystroke. Removing the second call means caching,
and its docs make "built fresh every call so what's drawn and what `⏎`/`↑`/`↓` act on can never
disagree" a deliberate correctness choice — not worth trading for ~0.9ms now that a build is cheap.

The lag and key-repeat symptoms were diagnosed and fixed by measurement, but **not confirmed
resolved in the running app** — that needs a real session on a large repo, which pairs naturally
with the skipped `verify` pass.

### DIFF
--- a/crates/app/src/palette/render.rs
+++ b/crates/app/src/palette/render.rs
@@ -376,6 +376,12 @@ impl AdeApp {
         }
         let next = (self.palette_selected as i32 + delta).clamp(0, total as i32 - 1);
         self.palette_selected = next as usize;
+        // Moving the highlight has to move the viewport with it, or the selection walks off the
+        // bottom and `⏎` runs a row the user cannot see (GitHub issue #413). Resolved on the next
+        // prepaint against real child bounds, and a no-op when the row is already fully visible.
+        if let Some(child) = palette::row_child_index(&groups, self.palette_selected) {
+            self.palette_results_scroll_handle.scroll_to_item(child);
+        }
         cx.notify();
     }
 
@@ -839,6 +845,9 @@ impl AdeApp {
 
         let mut container = div()
             .id("palette-groups")
+            // Test-only, as above: the results viewport's own painted box, so a test can assert a
+            // row is inside it rather than re-deriving the viewport from theme constants.
+            .debug_selector(|| "palette-results".to_string())
             .flex_1()
             .min_h_0()
             .overflow_y_scroll()
@@ -847,9 +856,18 @@ impl AdeApp {
             .flex_col()
             .py(px(4.0));
 
+        // Headers and rows are children of the scroller itself rather than of a per-group wrapper,
+        // so `palette_results_scroll_handle`'s child bounds line up with
+        // `palette::row_child_index` and a row can actually be scrolled to. The old wrapper was a
+        // bare `flex_col` with no styling of its own, so this lays out identically.
         let mut flat_index = 0usize;
         for group in groups {
-            container = container.child(self.render_palette_group(group, &mut flat_index, cx));
+            container = container.child(self.render_palette_group_header(group));
+            for entry in &group.entries {
+                let index = flat_index;
+                flat_index += 1;
+                container = container.child(self.render_palette_row(entry, index, cx));
+            }
         }
 
         // See `crate::sidebar::render::AdeApp::render_file_tree`'s own docs on why the scrollbar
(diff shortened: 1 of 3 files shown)


ColinEspinas:
- I don't like the "compacting effect" this adds.
- It seems the palette is laggy with a lot of elements (missing virtualized list ?).
- When keeping the up or down button pressed the selection stops moving and teleports once the key is released.

Issue #452, opened 2026-08-17
Standard commands like ls / cd / clean / etc. Do not work in the terminal


ColinEspinas:
## What happened

When typing standard commands like ls / cd / clean or other similar built in commands we have command not found messages.

## Where

Terminal

## Reproduction

Type clean in terminal inside jerry

## Environment

- OS: Windows 11
- Jerry version / commit: 0.1.1

ColinEspinas:
## Approach

**Root cause.** `TerminalSpec::default_shell_program()` (`crates/app/src/terminal/pane.rs:190`)
resolves the Windows default shell from `%COMSPEC%`, falling back to `cmd.exe`. Windows always
sets `%COMSPEC%`, and always to `C:\Windows\System32\cmd.exe`, so on Windows a Shell tab with no
override in Settings is *always* `cmd.exe`. Verified locally:

```
> & $env:COMSPEC /c ls
'ls' is not recognized as an internal or external command...
> & $env:COMSPEC /c clear
'clear' is not recognized as an internal or external command...
```

So `ls`, `clear`, `cat`, `pwd`, `rm`, `cp`, `mv` all fail. (`cd` *is* a `cmd.exe` builtin and does
work — the report groups it with the rest, but it isn't part of the bug.) This came in with #50,
which correctly fixed "Windows was trying to launch `/bin/bash`" but picked `%COMSPEC%` as the
Windows analogue of `$SHELL`. It isn't one: `%COMSPEC%` is documented as the *batch* interpreter,
not the user's preferred interactive shell.

**Fix.** Default to Windows PowerShell (`powershell.exe`) on Windows and stop consulting
`%COMSPEC%` for this decision. `powershell.exe` ships in System32 on every supported Windows and
is always on `%PATH%` — the same "always present" guarantee that made `cmd.exe` a safe fallback —
and it aliases `ls`, `cd`, `clear`, `cat`, `pwd`, `rm`, `cp`, `mv`, which is exactly what this
issue asks for. Pass `-NoLogo` on the default spawn so a small pane doesn't open on a copyright
banner; a shell the user typed into the Settings field is still launched verbatim with no
arguments, as today.

**Not preferring `pwsh.exe` (PowerShell 7) when installed**, deliberately: that needs a real
`PATH` search, and `default_shell_program_display()` is called from
`render_settings_shell_control` (`crates/app/src/settings/render.rs:1027`) on every frame — a
[...]

Issue #213, opened 2026-08-10
Allow to select shell


ColinEspinas:
For example on windows we want to be able to use bash.

Adding a setting to select which shell to use would be useful on all OS.

PR #459, opened 2026-08-17
fix(app): default a Windows terminal tab to PowerShell, not cmd.exe (#452)

ColinEspinas:
## What

A Windows terminal tab with nothing configured in Settings now runs Windows PowerShell instead of
`cmd.exe`, so `ls`, `clear`, `cat`, `pwd`, `rm`, `cp` and `mv` work — the commands #452 reports
failing. `-NoLogo` so a new pane doesn't open on PowerShell's copyright banner. A shell the user
names in Settings is still launched verbatim, with no arguments, exactly as before.

The Settings shell-suggestion list now leads with the probed shells rather than `%COMSPEC%`, so its
first entry is again the program an empty field actually runs. `cmd.exe` is still offered — it's a
real installed shell, just no longer the default.

Unix is untouched: `$SHELL`, falling back to `/bin/bash`, no arguments.

## Why

`TerminalSpec::default_shell_program` read `%COMSPEC%`, which Windows always sets, and always to
`cmd.exe`. So the fallback wasn't a fallback — it was every Windows session without an override.
`%COMSPEC%` names the *batch* interpreter, not an interactive shell, so it was never the Windows
analogue of `$SHELL` that #50 took it for.

Deliberately no `PATH` probe to prefer PowerShell 7's `pwsh.exe`: `default_shell_program_display`
feeds a Settings placeholder recomputed every frame, and a not-found `resolve_on_path` walks all of
`%PATH%` with no early exit — which is why the neighbouring detections in `settings/render.rs` are
already offloaded off the foreground thread. `pwsh` stays one click away in the auto-detected
suggestion list (#213).

## Testing

- [ ] `/check` — `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -D
      warnings` both pass. `cargo nextest run --workspace` is **not green on this Windows 11 host**,
      but not because of this change: 177 tests fail on a clean tree here (`pty-core
      pty_session_tests`, `lsp-core client::tests`, `merge::flow`, `review::render`, … — none of
      which this diff touches). I ran the full suite before and after and diffed the failing sets:
      **one** regression, `a_shell_never_receives_the_hook_injection_even_when_one_is_offered`,
      now fixed and passing. Zero other differences in either direction. CI's Linux and macOS jobs
      are the real signal for the suite.
- [x] New/changed behavior has a real test (`unit` tier)
  - `the_windows_default_shell_is_powershell_never_cmd_and_never_a_unix_path` — not `#[cfg]`-gated,
    so #452's guard (never `cmd.exe`) and #50's (never a Unix path) both run on every host,
    including CI's Linux job.
  - `an_unconfigured_windows_shell_is_powershell_started_without_its_banner` and
    `the_default_windows_shell_really_spawns` — `#[cfg(windows)]`, the second one spawning a real
    live PowerShell child over a real pty. Note CI's Windows job is build+clippy only (its
    `cargo test` is narrowed to `status_bar::process_stats`, see #440), so these ran on a real
    Windows 11 host rather than in CI.
  - `a_shell_never_receives_the_hook_injection_even_when_one_is_offered` asserted
    `shell.args.is_empty()`, which `-NoLogo` broke. Its real intent is "no *injected* args reach a
    shell", so it now asserts the shell's spec is identical with and without an injection on
    offer — a stronger check than the original, and platform-independent.
  - `windows_offers_real_path_hits_then_comspec_and_nothing_else` pins the new suggestion order.
- [ ] `verify` capture — not run: this touches no `render.rs`/`theme.rs`/layout, and the behaviour
      it changes is which process a pane spawns, which the real-pty test above covers directly.
- [ ] `external` tier — not touched.

### DIFF
--- a/crates/app/src/settings/state.rs
+++ b/crates/app/src/settings/state.rs
@@ -650,7 +650,8 @@ pub fn filter_keybinding_rows<'a>(
 /// (GitHub issue #213) - see [`detect_shell_status`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ShellStatus {
-    /// Nothing configured: a shell tab launches the real OS default (`$SHELL`/`%COMSPEC%`).
+    /// Nothing configured: a shell tab launches the real OS default (`$SHELL` on Unix,
+    /// `crate::terminal::pane::WINDOWS_DEFAULT_SHELL` on Windows).
     SystemDefault,
     /// A real program was found - the resolved absolute path, exactly as it would be run.
     Resolved(PathBuf),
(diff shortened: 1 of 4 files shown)


ColinEspinas:
I think this does not answer the issue at all as the initial report where using bash on windows.

Issue #176, opened 2026-08-05
Context-menu / Dropdowns should close when they lose focus


ColinEspinas:
When a dropdown loses focus it should be closed.

Example of not working today: 2 context menus open at the same time.

Also the commit one is particularly bugged and hard to close.

PR #464, opened 2026-08-17
feat(app): support the Cursor agent CLI, and let the start-agent button pick the agent

### LINKED ISSUE #463: feat(app): support the Cursor agent CLI, and let the start-agent button choose which agent to spawn
## What

Add the Cursor agent CLI (`cursor-agent`) as a third spawnable agent alongside Claude Code and Codex, and make the agent choice explicit at the point of starting a session instead of implicit.

Today `AgentKind` has exactly two variants (`Claude`, `Codex`, `crates/app/src/work_surface/agents.rs:17`) and every "start an agent" door resolves the kind *for* the user: `AdeApp::new_agent_pane` (`crates/app/src/work_surface/render.rs:1505`) runs a background `$PATH` search and spawns the first `settings::AGENT_KINDS` entry it finds installed. So on a machine with both `claude` and `codex` installed, there is no way to start a Codex session from the `Start an agent` button, the `+` menu's `New agent` row, the title bar's `New Agent Pane`, or `mod+shift+N` — only the command palette's separate `New Claude Agent` / `New Codex Agent` rows can pick. Adding a third agent makes that worse, not better.

Two changes, one issue because the second is what makes the first usable:

1. **`AgentKind::Cursor`** — spawns the `cursor-agent` binary from `$PATH`, exactly the way `Claude`/`Codex` already do (`ProcessKind::spec` → `TerminalSpec::command_with_env`). It appears in `settings::AGENT_KINDS`, so the Settings › Agents › Installed card gets a real third row with a live `$PATH`-derived ready/not-installed status, with no new detection code.
2. **An agent picker on the start-agent affordance** — `render_start_agent_button` (`crates/app/src/work_surface/render.rs:1913`) becomes a split control: clicking the button body keeps today's behaviour (spawn the resolved default, so no existing muscle memory or keybinding changes), and a caret opens a menu listing every `AGENT_KINDS` entry with its real installed state. Picking one spawns that kind. Same menu from the `+` menu's `New agent` row.

## Why

Cursor's agent CLI is the third of the three coding-agent CLIs people actually run in a worktree, and Jerry's whole premise is supervising several at once — being unable to start one of them is a real gap. And the picker is the part that turns "supported" into "usable": the current first-installed-wins resolution means the second and third agents are only reachable through the command palette, which is not where the empty pane's own `Start an agent` call-to-action points.

## Scope

Adding a variant to `AgentKind` is a deliberately loud change — it breaks every exhaustive `match`, which is the point. The real sites:

- `crates/app/src/work_surface/agents.rs` — `label`, `from_label`, `binary_name`, `ProcessKind::cursor()` constructor, and the existing per-kind tests.
- `crates/app/src/settings/state.rs:182` — `AGENT_KINDS: [AgentKind; 3]`.
- `crates/app/src/work_surface/state.rs` — `agent_tint`, `agent_initial`, `agent_icon_name`. Needs a tint from `theme::agent::TINT_POOL` that doesn't collide with a reserved hue (see `theme.rs`'s allocation rule and its own tests) and an icon-pack name (`"cursor"`).
- `crates/app/src/budget/state.rs:87` — `provider_of` returns `None` for Cursor. Honest, not a stub: the budget popover fetches real Claude/Codex quota endpoints, and there is no Cursor equivalent wired up. A fabricated budget row would be exactly the fake functionality CLAUDE.md forbids.
- `crates/app/src/palette/state.rs` — a `NewCursorAgent` command, matching the two that exist.
- `crates/app/src/run_history/model.rs:520` — the replay command line. `--resume` stays Claude-only (`hooks/flow.rs:58`); Cursor's own resume semantics are not verified here and won't be guessed at.
- Hook injection stays Claude-only too (`agents.rs:324` already gates on `AgentKind::Claude`), so a Cursor pane reports no hook status — same as Codex does today.

## Verification

`cursor-agent` is the binary name Cursor's CLI installer puts on `$PATH`. It is **not** installed on the machine this was written on, so before merge the spawn path needs one real check against an installed `cursor-agent`: that a bare invocation in a worktree starts an interactive session (the same thing `claude`/`codex` do), and that the Settings › Agents row resolves its path. If the real binary name differs, `binary_name` is the single line that changes.

## Design

No revision in `design_handoff_jerry_ade/` covers a split start-agent button. The menu reuses the existing dropdown row primitives (`render_dropdown_menu_row`, `theme::shadow::MENU`, `theme::zone::PLUS_MENU_WIDTH`) and the `+` menu's own open/close state pattern, so this stays inside the established visual language rather than inventing a new one.

lucasboinet:
Closes #463

## What

Two things, one PR because the second is what makes the first usable.

**`AgentKind::Cursor`** — spawns the `cursor-agent` binary from `$PATH`, exactly the way `Claude`/`Codex` already do. It joins `settings::AGENT_KINDS`, so Settings › Agents gets a real third row with a live PATH-derived status and no new detection code. Adding the variant breaks every exhaustive `match` on purpose; each one is handled explicitly rather than wildcarded.

**An agent picker.** Before this, every button door — the `Start an agent` CTA, the `+` menu's `New agent` row, the title bar's `New Agent Pane`, `mod+shift+N` — resolved the kind *for* the user with a first-installed-wins `$PATH` search. On a machine with two CLIs installed, the second was reachable only through the command palette. `Start an agent` is now a split button: clicking the body does exactly what it did before (so no muscle memory or keybinding changes), and a caret opens a picker listing every `AGENT_KINDS` entry with its real installed state. The `+` menu's `New agent` row opens the same picker instead of spawning outright.

The picker is registered as `MenuSurface::AgentPicker`, so it obeys issue #176's one-menu-at-a-time invariant in both directions.

Kept honest rather than stubbed:

- `budget::provider_of` returns `None` for Cursor — the popover's numbers come from `budget::fetch`'s real Claude/Codex endpoints and there is no Cursor equivalent wired up. A third row there would be fabricated.
- Hook injection and `--resume` stay Claude-only (`agents.rs` already gated on `AgentKind::Claude`), so a Cursor run restores as a fresh spawn carrying the same explicit "no resumable session id" notice Codex already gets.
- A picker row whose `$PATH` search really came back negative is disabled and says `cursor-agent not on PATH`; a search that *hasn't answered yet* stays clickable rather than being drawn as "not installed" — those are different facts.

## Why

Cursor's CLI is the third of the three coding-agent CLIs people actually run in a worktree, and supervising several at once is the app's whole premise. The picker is the part that turns "supported" into "reachable": the empty pane's own `Start an agent` CTA pointed at a door that could only ever open one of them.

## Testing

- [x] conventions ratchet (300/300 globs, 170/170 render adapter calls — both at baseline, unchanged)
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo nextest run --workspace` — **not run: `cargo-nextest` isn't installed on this machine.** A full `cargo test --workspace` was run instead and caught a real bug in an earlier revision of this branch (a renamed `agent.local.*` theme token failed every built-in theme file's validation, which is fixed here by keeping the public key name). A second full run against the final tree was still building when this PR was opened — I'll comment with the result rather than leave this box silently ticked.
- [x] New behaviour has real tests, in the `unit`/`ui` tiers:
  - `work_surface::state` — picker secondary text is three-state (ready / not-on-PATH / not-searched-yet), row-enabled rule, and no two `ProcessKind`s share a badge glyph or icon name.
  - `work_surface::render::agent_picker_tests` — a real click on the caret opens the picker anchored to that button and lists every kind; picking `Cursor` really spawns a Cursor agent; a kind the `$PATH` search says is missing can't be picked; the `+` menu's `New agent` row opens the picker instead of spawning.
  - `work_surface::agents` / `settings::state` — label⇄`from_label` round-trip, pairwise-distinct binary names, and that Cursor resolves `cursor-agent` rather than `cursor` (the latter is the editor's launcher, which would report an agent as ready that this app cannot spawn).
- [x] `verify` capture — **not attached: no release build / live window was run in this session.**

## Screenshot

Not captured — see the Testing note above. The picker reuses the existing dropdown primitives (`render_dropdown_menu_row`, `theme::shadow::MENU`, `theme::zone::PLUS_MENU_WIDTH`) and the `+` menu's own open/close and anchoring pattern rather than introducing new visual language, but this does want a real look before merge.

## Architecture notes

No crate-boundary change. One new piece of pure state in the GPUI-free `work_surface::state` (`AgentPickerAnchor` plus the two picker-row helpers), with the rendering and click wiring one layer up in `work_surface::render`, matching that module's existing split.

The steel-blue agent tint keeps its `agent.local.*` token key even though Cursor now owns the hue: that key is the public override every built-in and user theme already writes (`assets/themes/*.toml`'s `[agent] local.fg`), so renaming it would break theme files for a cosmetic gain. The built-in themes' inline comments are updated to name Cursor.

## Verification still needed

`cursor-agent` is the binary name Cursor's CLI installer puts on `$PATH`, but it is **not installed on the machine this was written on**. Before merge this wants one real check against an installed `cursor-agent`: that a bare invocation in a worktree starts an interactive session, and that the Settings › Agents row resolves its path. If the real name differs, `AgentKind::binary_name` is the single line that changes.

### DIFF
--- a/crates/app/src/run_history/model.rs
+++ b/crates/app/src/run_history/model.rs
@@ -8,7 +8,6 @@ use crate::hooks::history::{PastAgent, RunDiffstat};
 use crate::rail::status::Status;
 use crate::root::plural;
 use crate::theme;
-use crate::work_surface::agents::AgentKind;
 
 /// How a run ended.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -514,14 +513,17 @@ impl TranscriptLine {
 /// The command a resume of this run would really be - the transcript's opening line when none was
 /// captured. Only ever this run's own kind and its own checkout.
 fn resume_command_line(run: &PastAgent, branch: Option<&str>) -> String {
-    let command = match (run.kind, run.session_id.as_deref()) {
-        (AgentKind::Claude, Some(session_id)) => format!("claude --resume {session_id}"),
-        (AgentKind::Claude, None) => "cursor-agent".to_string(),
-        (AgentKind::Cursor, Some(session_id)) => format!("cursor-agent --resume={session_id}"),
-        (AgentKind::Cursor, None) => "cursor-agent".to_string(),
-        // Neither of these resumes: `Agents::spawn_resume` is Claude-only, so printing a
-        // `--resume` flag here would advertise something this app doesn't actually do for them.
-        (AgentKind::Codex, _) => "codex".to_string(),
+    // Built from the same `resume_args` the spawn itself uses, so this line cannot advertise a
+    // flag the resume would not really pass - a kind with no verified resume flag, or a run with
+    // no captured id, prints the bare binary and really does start fresh.
+    let binary = run.kind.binary_name();
+    let command = match run
+        .session_id
+        .as_deref()
+        .and_then(|session_id| run.kind.resume_args(session_id))
+    {
+        Some(args) => format!("{binary} {}", args.join(" ")),
+        None => binary.to_string(),
     };
     match branch {
         Some(branch) => format!("\u{276f} {command} \u{b7} {branch}"),
@@ -626,6 +628,7 @@ pub fn closing_lines(run: &PastAgent, now_unix: i64) -> Vec<TranscriptLine> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::work_surface::agents::AgentKind;
 
     fn run(worktree: &str, spawned: i64) -> PastAgent {
         PastAgent {
--- a/crates/app/src/run_history/model.rs
+++ b/crates/app/src/run_history/model.rs
@@ -516,11 +516,12 @@ impl TranscriptLine {
 fn resume_command_line(run: &PastAgent, branch: Option<&str>) -> String {
     let command = match (run.kind, run.session_id.as_deref()) {
         (AgentKind::Claude, Some(session_id)) => format!("claude --resume {session_id}"),
-        (AgentKind::Claude, None) => "claude".to_string(),
+        (AgentKind::Claude, None) => "cursor-agent".to_string(),
+        (AgentKind::Cursor, Some(session_id)) => format!("cursor-agent --resume={session_id}"),
+        (AgentKind::Cursor, None) => "cursor-agent".to_string(),
         // Neither of these resumes: `Agents::spawn_resume` is Claude-only, so printing a
         // `--resume` flag here would advertise something this app doesn't actually do for them.
         (AgentKind::Codex, _) => "codex".to_string(),
-        (AgentKind::Cursor, _) => "cursor-agent".to_string(),
     };
     match branch {
         Some(branch) => format!("\u{276f} {command} \u{b7} {branch}"),
--- a/crates/app/src/run_history/model.rs
+++ b/crates/app/src/run_history/model.rs
@@ -517,7 +517,10 @@ fn resume_command_line(run: &PastAgent, branch: Option<&str>) -> String {
     let command = match (run.kind, run.session_id.as_deref()) {
         (AgentKind::Claude, Some(session_id)) => format!("claude --resume {session_id}"),
         (AgentKind::Claude, None) => "claude".to_string(),
+        // Neither of these resumes: `Agents::spawn_resume` is Claude-only, so printing a
+        // `--resume` flag here would advertise something this app doesn't actually do for them.
         (AgentKind::Codex, _) => "codex".to_string(),
+        (AgentKind::Cursor, _) => "cursor-agent".to_string(),
     };
     match branch {
         Some(branch) => format!("\u{276f} {command} \u{b7} {branch}"),
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ actually done. Rebase interactively when you want to tidy up before merging.
 <p>
   <a href="https://claude.com/claude-code"><kbd><img src="https://www.google.com/s2/favicons?domain=claude.com&sz=64" alt="" width="16" valign="middle" /> Claude Code <img src="docs/images/dot-full.png" alt="Full support" width="9" valign="middle" /></kbd></a> &nbsp;
   <a href="https://github.com/openai/codex"><kbd><img src="https://www.google.com/s2/favicons?domain=openai.com&sz=64" alt="" width="16" valign="middle" /> Codex <img src="docs/images/dot-partial.png" alt="Partial support" width="9" valign="middle" /></kbd></a> &nbsp;
-  <a href="https://cursor.com/cli"><kbd><img src="https://www.google.com/s2/favicons?domain=cursor.com&sz=64" alt="" width="16" valign="middle" /> Cursor <img src="docs/images/dot-planned.png" alt="Planned" width="9" valign="middle" /></kbd></a>
+  <a href="https://cursor.com/cli"><kbd><img src="https://www.google.com/s2/favicons?domain=cursor.com&sz=64" alt="" width="16" valign="middle" /> Cursor <img src="docs/images/dot-partial.png" alt="Partial support" width="9" valign="middle" /></kbd></a>
 </p>
 
 <p>
--- a/assets/themes/ember.toml
+++ b/assets/themes/ember.toml
@@ -185,8 +185,8 @@ repo_fail_count    = "#984d60"  # The repo header's red urgency **count**
 "codex.bg"  = "#11291f"  # Teal - gpt-5-codex.
 "haiku.fg"  = "#606dab"  # Periwinkle - haiku-4.5.
 "haiku.bg"  = "#1a1d30"  # Periwinkle - haiku-4.5.
-"local.fg"  = "#4277a3"  # Steel blue - qwen-local.
-"local.bg"  = "#102538"  # Steel blue - qwen-local.
+"local.fg"  = "#4277a3"  # Steel blue - the Cursor agent CLI.
+"local.bg"  = "#102538"  # Steel blue - the Cursor agent CLI.
 
 # The Changes panel - four stacked, collapsible sections in one scroller, plus the file row's
 # own seen/unseen and discard states.
--- a/assets/themes/jerry-dim.toml
+++ b/assets/themes/jerry-dim.toml
@@ -185,8 +185,8 @@ repo_fail_count    = "#b46d61"  # The repo header's red urgency **count**
 "codex.bg"  = "#1a3533"  # Teal - gpt-5-codex.
 "haiku.fg"  = "#9683bf"  # Periwinkle - haiku-4.5.
 "haiku.bg"  = "#2c2537"  # Periwinkle - haiku-4.5.
-"local.fg"  = "#7c8dc2"  # Steel blue - qwen-local.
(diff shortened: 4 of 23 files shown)


crates/app/src/run_history/model.rs
```diff
@@ -517,7 +517,10 @@ fn resume_command_line(run: &PastAgent, branch: Option<&str>) -> String {
     let command = match (run.kind, run.session_id.as_deref()) {
         (AgentKind::Claude, Some(session_id)) => format!("claude --resume {session_id}"),
         (AgentKind::Claude, None) => "claude".to_string(),
+        // Neither of these resumes: `Agents::spawn_resume` is Claude-only, so printing a
+        // `--resume` flag here would advertise something this app doesn't actually do for them.
         (AgentKind::Codex, _) => "codex".to_string(),
+        (AgentKind::Cursor, _) => "cursor-agent".to_string(),
```
ColinEspinas: https://cursor.com/docs/cli/using#history ?

lucasboinet:
## `verify` run — real app, real `$PATH`

Ran the `verify` loop against a debug build of this branch (`cargo build -p app`, launched on this worktree). The skill is written for macOS (`osascript` + `screencapture`); on Windows I used the equivalent Win32 calls — `GetWindowRect` for the window bounds, `Graphics.CopyFromScreen` for the region capture, and `SetCursorPos`/`mouse_event` for the clicks — so the UI was driven with real mouse input, not simulated events inside a test context.

What the running app showed:

**1. The split button renders as designed.** The agent context bar's CTA reads `Start an agent · Ctrl Shift N · ▾`, with the caret as a separate hit area behind a 1px divider.

**2. A real click on the caret opens the picker, anchored under the button**, listing all three kinds with live `$PATH` state. On this machine only `claude` is installed, and the picker said exactly that:

- `Claude` — copper badge, `runs in master`, enabled
- `Codex` — dimmed badge and label, `codex not on PATH`, not clickable
- `Cursor` — dimmed badge and label, `cursor-agent not on PATH`, not clickable

That's the honest-detection path proving itself against a real filesystem rather than a fixture: the two uninstalled CLIs are listed (so the feature is discoverable) but can't be picked, and each says which binary is missing. Note it names `cursor-agent`, not `cursor`.

**3. The `+` menu's `New agent` row opens the same picker**, anchored to the `+` button, with the `+` menu itself closing as it opens — the one-menu-at-a-time invariant holding across the two surfaces.

**4. Picking a row really spawns that kind.** Clicking `Claude` opened a new agent tab, the rail row went to `running · Claude`, the status pill to `Running`, and the pane showed a live `claude` process (pid 28524, Claude Code v2.1.234) in the worktree.

No mismatches found against the issue's acceptance criteria. The design revision set (`docs/design/`, formerly `design_handoff_jerry_ade/`) has no entry covering a split start-agent button, so the issue was the oracle here; the popover reuses `render_dropdown_menu_row`, `theme::shadow::MENU` and `theme::zone::PLUS_MENU_WIDTH`, so it matches the `+` menu it sits beside.

**The PNGs are not attached here**: `gh` has no image-upload path, and `.claude/scratch/` is gitignored precisely so captures never land in a commit. They're on the branch author's machine at `.claude/scratch/0{1..5}-*.png` and can be dragged into this PR if a visual record is wanted in-thread.

## Test suite — still blocked, and not by this change

`cargo nextest` isn't installed here, and the fallback `cargo test --workspace` run died in the linker with a full disk:

```
rustc-LLVM ERROR: IO failure on output stream: no space on device
app-...exp : fatal error LNK1108: impossible d'écrire le fichier à 0x0
```

So the behavioural tests added in this PR have compiled but have not been *run* on the final tree. `fmt --check`, `clippy --workspace --all-targets -D warnings`, and the conventions ratchet (300/300, 170/170) all pass. CI is the real check here until there's disk headroom locally.

ColinEspinas (commented):
No hooks like claude ? This seems like half the implementation

ColinEspinas:
Still missing hooks like claude. Cursor should have a similar system.

lucasboinet:
## Why the Cursor agent gets no hooks

Short version: not because `cursor-agent` lacks hooks, but because it has no way to be *told where its hook config lives* — and every available location writes into something that isn't ours.

### Cursor does have a hook system

Verified against `cursor-agent 2026.08.11-e8db854`. It ships 18 lifecycle events, and every hook payload carries a common set of base fields including `conversation_id`, `workspace_roots` and `transcript_path`. On paper Jerry's whole side-channel ports: the status facts, the session id, run titles/turn counts, and the file-edit stream `provenance` attributes lines with.

Even the transport survives. Hooks run as child processes, so the `JERRY_HOOK_PORT` / `JERRY_HOOK_TOKEN` / `JERRY_AGENT_ID` environment Jerry injects on the pane propagates to them, and the forwarder script would work unchanged.

### What actually blocks it

Jerry's Claude integration hangs entirely on one flag:

```
claude --settings <generated file in Jerry's own temp dir>
```

That is what makes the whole mechanism zero-footprint. The settings file and forwarder script live in a per-launch `jerry-hooks-<pid>-<random>/` directory (0700 on unix), and dropping the `HookRuntime` deletes them. Nothing is written into the user's repo, and nothing is written into `~/.claude`.

**Cursor has no equivalent flag or environment variable.** Hooks load from exactly four fixed locations, in priority order:

1. Enterprise (MDM-managed, system-wide)
2. Team (cloud-distributed, Enterprise only)
3. Project — `<project-root>/.cursor/hooks.json`
4. User — `~/.cursor/hooks.json`

So Jerry would have to write into one of two places it does not own:

- **The worktree** (`<worktree>/.cursor/hooks.json`) — a real file in the user's repository. It shows up in Jerry's own Changes pane and review surface, it lands in the diff the agent is being reviewed on, and the agent itself can commit it. Jerry would be dirtying the very worktree whose diff is the product.
- **`~/.cursor/hooks.json`** — changes the user's Cursor everywhere, including outside Jerry entirely, and two Jerry instances would contend over a single global file.

Neither is a thing to do quietly as a side effect of clicking "start an agent". That is a product decision, so it is deliberately not made here.

### Secondary mismatches

Real, but none of them are the blocker:

- **Different schema.** Claude nests as `hooks.Stop[0].hooks[0].command`; Cursor is `{version: 1, hooks: {stop: [{command}]}}`.
- **Different event names**, and not a clean 1:1 — Jerry forwards 9 Claude events, and Cursor has no equivalent of `Notification` or `StopFailure`. The `NeedsInput` fact, which drives the "blocked on you" state in the rail, would have to be re-derived from a different signal.

### What is lost, concretely

- **Live status** falls back to the terminal-title and quiescence signals — the same fallback that already runs when the hook listener can't start. Status still works, just less precisely.
- **No run-history entry.** `record_agent_statuses` records only agents with a fresh hook signal, deliberately: the store refuses to persist a status inferred from pty silence, because a guess written to disk is still a guess an hour later. No hookless kind is ever persisted, so Cursor runs will not appear under History and its "Resume here" will never list one.

### What is not lost

Resume. That is what this PR adds, and it works without hooks: `cursor-agent create-chat` prints a chat id up front, so Jerry mints one and starts the agent attached to it, rather than waiting for a hook to report one. Cursor sessions are therefore resumable through tab-session restore across an app restart.

Reopening this later is a matter of deciding where that file may be written, not of new capability in the CLI.

